PR #1, opened 2015-06-30
Upcoming elections handler

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/95986678)

Adds a message handler listening to `electorate.create` events, and creating one or more upcoming-election notifications if applicable.

Will be changing somewhat once we have scheduling worked in.

Assigning to @tie-rack


src/election_notification_works/channels.clj
```diff
@@ -1,8 +1,38 @@
...
+                                   (if (nil? event)
+                                     :closed
+                                     (do
+                                      (try
+                                        (handler event)
+                                        (catch Throwable t
+                                          (log/error t "Error handling event." (pr-str event))))
+                                      (recur)))))))]
+    {:stop! (fn [] (async/put! stop-chan :stop))
+     :notes signal-chan}))
```
tie-rack: This map `:stop!` and `:notes` isn't used anywhere. Why is the `stop-chan` needed? Couldn't this be a go-loop that stops if the channel is closed? What does `:notes` mean in this context anyway?

src/election_notification_works/entities/subscription.clj
```diff
@@ -0,0 +1,48 @@
+(ns election-notification-works.entities.subscription
+  (:require [datomic.api :as d]
+            [datomic-toolbox :as db]))
+
+(db/deftx-data-fn subscription->tx-data
+  [user-id mediums]
+  {:id user-id
+   :tx-data [{:db/id ?id
+              :voter/id user-id
+              :notification/medium mediums}]})
```
tie-rack: This function isn't exactly returning *just* tx-data. And the `:id` is never used when this function is called. Can it be simplified to only be the actual tx-data?

src/election_notification_works/entities/subscription.clj
```diff
@@ -0,0 +1,48 @@
...
+  [entity]
+  {:user-id (:voter/id entity)
+   :mediums (set (:notification/medium entity))})
+
+(defn read-subscription-entity
+  [user-id]
+  (db/find-one '{:find [?subscription]
+                 :in [$ ?user-id]
+                 :where [[?subscription :voter/id ?user-id]]}
+               (db/db) user-id))
```
tie-rack: We've generally been writing functions like this to take a database as an argument, so that it can be consistent within a set of operations.

src/election_notification_works/handlers.clj
```diff
@@ -1,8 +1,112 @@
...
+        cutoff (time/minus date (time/days 24))]
+    (time/after? now cutoff)))
+
+(defn subscription-for-user
+  "Retrieves the users' subscription."
+  [user]
+  (subscription/read-subscription (:id user)))
+
+(defn filter-subscription-mediums
+  "Takes in the full `subscription type and `user hash,
```
tie-rack: Backticks should surround what they're adorning.

src/election_notification_works/channels.clj
tank157: I'd like to tag @ericnormand on this one, it came from the work he and @bwreid started. I think it's there for some future back-pressure type work, but I'm not 100% sure on that.

src/election_notification_works/entities/subscription.clj
tank157: Ah, good catch, I'll update that.

src/election_notification_works/handlers.clj
```diff
@@ -1,8 +1,112 @@
...
+          :context electorate})
+       mediums))
+
+(defn send-notifications
+  "Puts each of the `notifications onto the `channel separately."
+  [channel notifications]
+  (async/go-loop [remaining notifications]
+    (when-let [message (first remaining)]
+      (async/>! channel message)
+      (recur (rest remaining)))))
```
tie-rack: Could the entire body of this function be `(async/onto-chan channel notifications false)`?

src/election_notification_works/channels.clj
ericnormand: This looks like it was copied from another repo. I noted in the PR for this change that this could be extracted into its own library. I added the stop! function to be able to stop the thread for any reason without closing the channel (which is presumably hooked up to a langhor queue). This is just a good idea generally when making threads since there's no other way to stop them. That said, if the channel is closed, this thread will stop. The notes channel may be poorly named. It's short for "notifications". It lets you know when and why the thread stopped.

`go-loop` is not appropriate since we're doing IO in the handler. That could gum up the go routines' thread pool.

src/election_notification_works/handlers.clj
ericnormand: Looks like it to me.

src/election_notification_works/channels.clj
tie-rack: If that map isn't being used (and the `!` in the function name sort of implies it's called just for side effects), I'd suggest getting rid of it (and the deal with `stop-chan` since it's also entirely unused).

ericnormand: The entire function is about to go away into a library. Please leave it the same so it's obvious that it can be replaced.

tank157: I'll leave this alone for now. Chris, would you be ok with this staying the same, seeing as it's already in other, accepted work and is slated to be refactored into a library?

As for the `:notes` channel, what if we dropped it and just responded on the `stop-chan` whenever and for whatever reason we stop, even if the signal came in on the stop-chan? Or is it generally frowned upon to use channels in "both directions" like that?

tie-rack: Sure. It can stay the same (though not because it was accepted in the past, I don't find that to be a compelling reason). But if you're arguing for it staying the same... why are you suggesting changing it?

tank157: The changes wouldn't be now, I was more thinking about the version that will go in a library.

tank157:
@tie-rack Can you re-review this with the latest changes?

tie-rack:
Awesome. LGTM.

PR #5, opened 2015-07-17
replace voter w/ user

cap10morgan:
...to be consistent with the rest of the kraken. We used voter in Classic and in Ballot Scout, but we should use "user" in the kraken. When/if we use "voter" in the future, it should mean that we're modeling something about being a voter, not just as a clever name for our users. In this case, one needn't be a voter to get election notifications, so user is more appropriate.

Just a small Friday change.


cap10morgan:
@danielglauser ping

danielglauser:
LGTM, ship it

PR #7, opened 2015-07-30
[TH] refactors handlers

tank157:
Now the main message type handlers are no longer
directly listening to the `electorate.create`, 
instead a single handler takes care of figuring
out subscriptions and validates them and then
calls each notification type handler as needed.

Based on a recommendation from Eric, will make
triggering notifications based on subscriptions
easier as well.

No Pivotal Card of this as of now.

Assigning to @ericnormand as it was inspired by him.


ericnormand:
LGTM

PR #10, opened 2015-08-12
Subscription confirmation email

tank157:
[Pivotal Card 1](https://www.pivotaltracker.com/story/show/99238726)
[Pivotal Card 2](https://www.pivotaltracker.com/story/show/100581522)

Notification types have been uploaded to production.
Upgraded to kehaar 0.5.0

Assigning to @cap10morgan


tank157:
Also, worked in krakenstein

src/election_notification_works/channels.clj
```diff
@@ -4,36 +4,25 @@
...
 (defonce electorate-create-events (async/chan))
+(defonce subscription-create-in (async/chan))
+(defonce subscription-delete-in (async/chan))
 
 ;; outgoing
+(defonce ok-out (async/chan))
 (defonce notification-create (async/chan))
 (defonce notification-schedule (async/chan))
+(defonce subscription-create-out (async/chan))
+(defonce subscription-delete-out (async/chan))
```
cap10morgan: I've been calling these types of channels `whatever-requests` and `whatever-responses` in my kehaar 0.5.0 works. Obviously those are longer but I think also a bit clearer about what they're intended for. Either way, I don't have a strong opinion on what we name them between your choice and mine, but it does seem like we should pick and document a convention sooner rather than later. Right?

src/election_notification_works/channels.clj
```diff
@@ -4,36 +4,25 @@
...
+
+;; external-service
+(defonce read-user (async/chan))
```
cap10morgan: Seems like we typically do `entity-action` for these names, don't we?

src/election_notification_works/channels.clj
tank157: I don't really have any strong feelings here myself, but wire-up uses incoming, outgoing, in-channel, out-channel, etc. So this is at least more in line with that.

cap10morgan: Yeah, that's true. But it's also a general purpose library that has to stay agnostic about why things are going in and out. I guess the thing I like most about `requests` and `responses` is that it differentiates those from incoming / outgoing event channels and outgoing external service channels. We may want some naming conventions around those too, I guess.

dcohen:
LGTM!

PR #12, opened 2015-08-24
Update deps

cap10morgan:
I was about to start a card in here, and as I am wont to do, thought I would update the dependencies first (as well as one other small fix in docker-compose.yml–making it expose various ports to the host on the ports we've begun standardizing on).

Tests still pass and it runs in docker-compose.


ericnormand:
LGTM

ericnormand:
Nice separation of work!

PR #13, opened 2015-08-25
Use preferred first name (nickname or first-name if no nickname)

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/101134418)

I TDD'd this. Tests pass and it runs in its docker-compose stack. Incoming RabbitMQ message expectations are unchanged with the exception of looking for `:nickname` in users. Outgoing RabbitMQ message schemas are unchanged.


tank157:
LGTM...ran tests, fired up in docker-compose

PR #14, opened 2015-08-27
Only confirm new subscriptions

cap10morgan:
turbovote-web users were seeing a confirmation on every step change because we're (unfortunately) re-PUT'ing the subscriptions on each one for now.

This makes subscription PUTs idempotent w/r/t confirmation messages.


src/election_notification_works/entities/subscription.clj
```diff
@@ -15,28 +15,27 @@
...
-  (some-> (read-subscription-entity db user-id)
-          entity->subscription))
+  ([user-id] (read-subscription (db/db) user-id))
+  ([db user-id]
+   (some-> (read-subscription-entity db user-id)
+           entity->subscription)))
 
 (defn update-subscription
   [old-subscription new-subscription]
-  (if (empty (set/difference (:mediums old-subscription)
```
ericnormand: Ah! I'm glad you got rid of this. Looks like a mistake between `empty` and `empty?`

cap10morgan: Exactly. You'll see that it's now using `empty?` in its new home in the `handlers` ns.

ericnormand: Yep!

ericnormand:
Nice tests.

ericnormand:
LGTM

cap10morgan:
@ericnormand Thanks for the fast after-hours review!

PR #17, opened 2015-09-07
Transaction function test

ericnormand:
To run this, you will need to install datomic-toolbox from [this branch](https://github.com/democracyworks/datomic-toolbox/tree/datomic-transaction-function). Use `lein install` to install it locally.

This branch modifies `election-notification-works.entities.subscription` to handle all datomic operations atomically. This was a good proof of the database functions since it has challenging semantics. It's not simple entity updates. It has to maintain subscription uniqueness (each userid can have at most one subscription) and it also has to remove the subscription if one of the properties has 0 elements (:notification/medium).

This required a little bit more work, since these are not modifications to existing entities. Rather, they are general conditions that required being able to send in an arbitrary query. Luckily, that seems to be as general as possible as well.

(Note that the tests will fail in BuildKite because the JAR is not available)


ericnormand:
Ping @cap10morgan to see what you think.

cap10morgan:
Reviewing this is going to require that I spend some quality time with the changes to datomic-toolbox first. But I should do that anyway. I may wait until it gets a LGTM from @tie-rack, though.

ericnormand:
Ping @cap10morgan the datomic-toolbox PR has been LGTM'd

cap10morgan:
@ericnormand The amount of work that has piled up on my plate between now and the end of the sprint probably calls for re-assigning this PR. But if I'm still as good as anyone else at this point, I will try to get to it.

src/election_notification_works/entities/subscription.clj
```diff
@@ -28,40 +31,54 @@
...
   [subscription]
-  (if-let [old-subscription (read-subscription-entity
-                             (db/db) (:user-id subscription))]
-           (update-subscription old-subscription subscription)
-           (let [{:keys [db-after]} @(db/transact
-                                      [(subscription->tx-data subscription)])]
-             (read-subscription db-after (:user-id subscription)))))
+  (let [userid (:user-id subscription)
+        {:keys [db-after db-before]}
+        @(db/swap-tx!
```
tank157: So am I right that if the TX fails here, presumably because the assert-empty, it will retry it 100 times?

ericnormand: Yep, if it fails with a ConcurrentModificationException, yes. Other failures fail.

tank157: 100 times feels like overkill for the create, but maybe not and you've thought of a reason why. I suppose it could be that we might have deletes and creates with the same user-id coming in close to each other?

src/election_notification_works/entities/subscription.clj
```diff
@@ -28,40 +31,54 @@
...
-        new-mediums (:mediums new-subscription)
-        txs (into [] (for [m new-mediums]
-                       [:db/add id :notification/medium m]))
-        {:keys [db-after]} @(db/transact txs)]
-    (read-subscription db-after (:user-id new-subscription))))
+        userid (:user/id old-subscription)
+        old-mediums (:notification/medium old-subscription)
+        new-mediums (:mediums new-subscription)]
+    ;; assert that we haven't lost/changed the subscription
+    [[:assert-not-empty
```
tank157: This seems like a double read to me. Which may not be actually bad, I'm just trying to reason out why not, and that there isn't a way around this with maybe `:assert-equal`, but perhaps that's just the same thing in a different style.

src/election_notification_works/entities/subscription.clj
```diff
@@ -28,40 +31,54 @@
...
-                                      [(subscription->tx-data subscription)])]
-             (read-subscription db-after (:user-id subscription)))))
+  (let [userid (:user-id subscription)
+        {:keys [db-after db-before]}
+        @(db/swap-tx!
+          (fn [db]
+            (if-let [old-sub (read-subscription-entity db userid)]
+              (update-subscription-tx old-sub subscription)
+              (create-subscription-tx subscription))))]
+    {:before (read-subscription db-before userid)
```
tank157: This use of `db-before` is so cool!

src/election_notification_works/entities/subscription.clj
tank157: I suppose it might just be a consequence of wanting the "create" to do an upsert. We have to read it at least once to know if we're creating something new or updating something existing.

src/election_notification_works/entities/subscription.clj
```diff
@@ -28,40 +31,54 @@
...
+  (let [{:keys [db-after]}
+        @(db/swap-tx!
+          (fn [db]
+            (when-let [old-sub (read-subscription-entity db user-id)]
+              (let [old-mediums (:notification/medium old-sub)
+                    id (:db/id old-sub)
+                    new-mediums (set/difference old-mediums mediums)]
+                (if (empty? new-mediums)
+                  [ ;; assert same subscription
+                   [:assert-not-empty
```
tank157: So similar to my other thought, could this not just be something like:

`[:assert-equal (:user/id old-sub) id :user/id]`

?

src/election_notification_works/entities/subscription.clj
ericnormand: 100 is meant to be overkill. It's supposed to be essentially infinite, while still being finite. I'm up for changing it but it's pretty arbitrary. 10 seems too low for a default.

src/election_notification_works/entities/subscription.clj
ericnormand: You mean the `:assert-not-empty` could be replaced by an `:assert-equal`? Could be. How would you suggest doing it?

src/election_notification_works/entities/subscription.clj
ericnormand: Yeah, that would be good. It may be that not-empty is unnecessary.

src/election_notification_works/entities/subscription.clj
tank157: I was thinking along the lines of:

`[[:assert-equals (:user/id old-subscription) id :user/id]
  [:transact id :notification/medium old-mediums
      (conj old-mediums new-mediums)]]`

src/election_notification_works/entities/subscription.clj
tank157: Perhaps in this case, but if we wanted to assert more than one attribute remaining constant, the `:assert-not-empty` makes for a way to do them all in a single line.

test/election_notification_works/entities/subscription_test.clj
```diff
@@ -0,0 +1,124 @@
...
+(use-fixtures :each with-fresh-db)
+
+(defn pauser [f ch]
+  (fn [& args]
+    (let [ret (apply f args)]
+      (async/<!! ch)
+      ret)))
+
+(deftest create-subscription-concurrent-test
+  (testing "create subscription with concurrent modification"
```
tank157: Maybe change the testing string here...otherwise it's the same one as the next test, which seems closer to concurrent modification to me.

test/election_notification_works/entities/subscription_test.clj
```diff
@@ -0,0 +1,124 @@
...
+                     (async/thread
+                       (async/<!! chan)
+                       (create-subscription {:user-id userid
+                                             :mediums #{:sms}}))))]
+      (async/close! chan)
+      (doseq [thread threads]
+        (async/<!! thread))
+      (let [results (d/q '{:find [?eid]
+                           :in [$ ?userid]
+                           :where [[?eid :user/id ?userid]]}
```
tank157: Could we add to the `:where` something like `[?eid :notification/medium #{:sms}]` just to make sure that we only get 1 subscription and the medium is correct?

src/election_notification_works/entities/subscription.clj
tank157: So does the retry attempt to apply the same tx again, or does it modify itself to new conditions? Maybe that's what's not clear to me. In the case of create, it's unlikely (unless a delete comes along) to ever successfully commit in those 100 retries.

tank157:
All in all, this seems really great and I can't want to try my hand at some of it!

src/election_notification_works/entities/subscription.clj
ericnormand: If there's a ConcurrentModificationException, the function you pass in is called again with the newest version of the db. So that will regenerate the transaction data that's sent to it based on the most recent db you've got.

In this case, the first time if you do a create-subscription-tx and it fails, the next time it should do an update-subscription-tx because that choice is based on the current db.

test/election_notification_works/entities/subscription_test.clj
ericnormand: Modified the test to check that there's one subscription and the mediums is #{:sms}.

tank157:
Changes look good, and my remaining questions aren't really enough to hold anything up, so LGTM

ericnormand:
Alright, this is tested with `lein run` and `docker-compose up` and with `krakenstein`. Take that!

Note that these tests *did* reveal bugs. So great!

tank157:
Nice! That sort of core bug is insidious, no? It works in test just fine, sure, but that's not under test.

PR #20, opened 2015-09-18
Add absentee ballot url to notification context.

ericnormand:
Pretty simple code. Just include the map in absentee_ballots.clj and look up the keywordized state from the map.

This is apparently a stopgap. These URLs will soon be in their own works component.

[Pivotal Card](https://www.pivotaltracker.com/story/show/102966516)


ericnormand:
Tested in Krakenstein.

ericnormand:
Also tested with `lein run` with new tests.

7maples:
LGTM

PR #21, opened 2015-09-21
Add same-day-voter-reg flag to context.

ericnormand:
Before this card was started, we had text in the template that *always* said "You can register on election day" in the upcoming election notification. This update adds a boolean to the context that says whether this is true. If it's not true, we don't show the text.

We determine if it's true by seeing if the district division in the electorate has a `:election-day` voter-registration-option. If it's there, we can show the text, otherwise, we can't.

This PR also involved changing the upcoming-election-body.hbs template. (https://github.com/democracyworks/notification-templates/pull/2)

[Pivotal Card](https://www.pivotaltracker.com/story/show/102029618)


bwreid:
:+1: Looks good to me.

cap10morgan:
Tested in krakenstein and works!

PR #22, opened 2015-09-22
Basic updates

cap10morgan:
In preparation for the "schedule upcoming-election notifications" work, I updated the dependencies and added the standard logging config. But this revealed a very strange bit of code that should never have worked, but was for some reason (calling `name` on a function, specifically). I have no idea why it was working before, but changing almost anything in the `:dependencies` map in `project.clj` made it stop working. Since it shouldn't have worked in the first place, I just removed it.


cap10morgan:
Merging this so it doesn't blow up on us while running notifications Monday morning.

PR #23, opened 2015-09-22
Schedule upcoming election notifications

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/99238048)

This schedules upcoming election notifications instead of sending them immediately.

It is based on the `basic-updates` branch so that those changes can be independently reviewed.


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -15,6 +15,41 @@
...
+  {:email [7 14]
+   :sms [12 42]})
+
+(defn voter-reg-by-mail-deadline
+  [election]
+  (let [district-division (:district-division election)
+        method (first (filter #(= (:type %) :by-mail)
+                              (:voter-registration-methods district-division)))]
+    (->> method
+         (#(or (:deadline.received %) (:deadline.postmarked %)))
```
danielglauser: Sounds like we should never have both `:deadline.received` and `:deadline.postparked` but if we do it looks like we prefer received. Is that correct behavior?

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -15,6 +15,41 @@
         cutoff (time/minus date (time/days 24))]
     (time/after? now cutoff)))
 
+(def medium->time
```
danielglauser: Should the hour/minute that we send both emails and SMS messages be `config`urable? No is a perfectly valid response, just checking.

src/election_notification_works/handlers/upcoming_election.clj
cap10morgan: Yes, we should never have both. If we do have both, behavior is unspecified.

src/election_notification_works/handlers/upcoming_election.clj
cap10morgan: Sure, that's not a bad idea.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -15,6 +15,41 @@
...
+
+(defn ideal-send-at
+  ([medium election] (ideal-send-at medium election util/NY))
+  ([medium election tz]
+   (let [election-date (coerce/from-date (:date election))
+         thirty-days-prior (time/minus election-date (time/days 30))
+         reg-deadline (voter-reg-by-mail-deadline election)
+         five-days-prior-to-reg-dl (time/minus reg-deadline (time/days 5))
+         send-time (partial medium-send-time medium tz)]
+     (->> [thirty-days-prior five-days-prior-to-reg-dl]
```
danielglauser: This reads really nicely

danielglauser:
Thanks for the walkthrough. LGTM

cap10morgan:
I tried testing this in krakenstein but that revealed that electorate-works is broken. I have created a [bug card](https://www.pivotaltracker.com/story/show/104043912) for that. This should not be merged until that is fixed and this is verified to work in krakenstein.

cap10morgan:
This is tested and working in krakenstein now.

PR #25, opened 2015-10-13
[TH] wrap tests with hardcoded dates with time/do-at so they don't sl…

tank157:
…ip into the past

Assigning to @ericnormand


ericnormand:
LGTM

PR #27, opened 2015-10-15
add polling-places/lookup fn

cap10morgan:
Uses Google's Civic Info API to lookup polling places by voter address
and election-id.

[Pivotal card](https://www.pivotaltracker.com/story/show/104476034)


danielglauser:
Pure gold here, LGTM

PR #28, opened 2015-10-15
Polling place in vote-tomorrow notification handler

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/104923672)

We're now looking up polling places via the Google Civic Info API in the vote-tomorrow handler. So any notifications that come through after polling places are available will just get them and everything will be so great!


src/election_notification_works/polling_places.clj
```diff
@@ -8,14 +8,17 @@
...
-                                             :state
-                                             :zip]))
-        voter-info-partial (partial gc/voter-info (config [:google-api-key])
-                                    addr-str)
-        voter-info (if google-election-id
-                     (voter-info-partial google-election-id)
-                     (voter-info-partial))]
-    (:pollingLocations voter-info)))
+  (let [api-key (config [:google-api-key] :disabled)]
+    (if (= :disabled api-key)
```
tie-rack: I have a slight preference for combining this `let` and `if` (and getting rid of the magic `:disabled` value that isn't really used):

```clj
(if-let [api-key (config [:google-api-key] nil)]
  ...)
```
And then switch the clauses. Though, from the look of it, `nil` would be an acceptable return value, so then just a `when-let`.

cap10morgan: We started with that, but the problem that led us to this is that returning `nil` as the default value here is indistinguishable from the env var not being set and using `#resource-config/env` in the `config.edn` (which we are). That leads to silent failures if you forget to set the env var.

cap10morgan:
@tie-rack Better?

src/election_notification_works/polling_places.clj
```diff
@@ -8,14 +8,18 @@
...
+                                               :state
+                                               :zip]))
+          voter-info-partial (partial gc/voter-info
+                                      api-key
+                                      addr-str)
+          voter-info (if google-election-id
+                       (voter-info-partial google-election-id)
+                       (voter-info-partial))]
+      (:pollingLocations voter-info))
+    []))
```
tie-rack: Is there a specific reason this is an empty vector? Or would `when` and letting this fn return `nil` be just as good?

cap10morgan: I like the empty vector because when the lookup is successful, this function returns a vector of polling places. So any condition where there are no polling places seemed like it should return an empty vector. But I don't feel *that* strongly about it. If you think returning an empty vector when the API call is successful but there are no polling places found but returning nil when API calls are disabled is better, I'd be happy to make it so.

tie-rack: No. It just seemed like over-specifying. But I guess it's fine.

cap10morgan:
@tie-rack OK. So... LGTY?

tie-rack:
Yeah. You got it.

PR #29, opened 2015-10-19
Un-hardcode voting

danielglauser:
Took out the hard-coded voting method and added in some code to transform the voting methods into something that Handlebars can work with. Added some convenient fns to `test-helpers.` Will supply the right context for [this](https://github.com/democracyworks/notification-templates/pull/5) commit.


7maples:
Is there a Pivotal card for these changes?

src/election_notification_works/handlers/util.clj
```diff
@@ -48,3 +48,18 @@
 (defn medium->time
   [notification-type medium]
   (config [:notification-types notification-type medium :send-at]))
+
+(defn user-district-division
+  "Determines which of the `election`'s district-divisions the `user` is
+  registered to vote in.
+
+  This fn should go away as soon as we make electorate-works remove the
```
7maples: Should this really be a code comment? Suggest adding a task to that other card.

cap10morgan: @7maples It is in a card (a few cards, actually). We're just trying to make it easier to find the places where we need to change the code once those cards are done.

cap10morgan:
@7maples No card. This is something we realized we'd forgotten about late last week but is needed for vote-tomorrow notifications to say the right things to everyone about voting ID requirements.

src/election_notification_works/handlers/util.clj
7maples: Right, I was thinking it would be easier for the person picking up the card you're pointing to here to have a task on that card unless they'll otherwise know to look at this and other comments. But that's fine, this gives people a trail back to the card. Code comment first though, for sure.

7maples:
LGTM

PR #30, opened 2015-10-22
Change vote tomorrow send-to queue

cap10morgan:
This schedules vote-tomorrow notifications to come back here so we can look up polling places right before they go out.

It is the second half of [this card](https://www.pivotaltracker.com/story/show/104927444).


danielglauser:
LGTM

PR #31, opened 2015-10-26
Vote tomorrow update

danielglauser:
Adds a handler to update the vote tomorrow notifications with their respective polling places. Ticket [here](https://www.pivotaltracker.com/story/show/104924120).


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -112,3 +114,13 @@
     (let [notifications (->vote-tomorrow-notification-messages mediums
                                                                electorate)]
       (util/send-notifications channel notifications))))
+
+(defn update
```
ericnormand: can we rename `update` to something that is more explanatory and doesn't conflict with a function in clojure.core?

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -108,7 +111,18 @@
...
+
+(defn handle-update
+  "Takes an incoming `notification`, looks up the polling locations
+  and adds one of them to the context. The `channel` is used to send out
+  the new notification."
+  [channel {:keys [recipient election] :as notification}]
+  (->> (assoc notification :polling-place
+              (user-polling-place recipient election))
+       (async/>!! channel))
+  channel)
```
ericnormand: I'm sorry I didn't see this before, but why are you returning the channel? That's the event channel, as I understand it. If you return a channel from a kehaar handler, it will consume the first thing put on the channel. Kehaar responder handlers should return some value, usually a map, with a status, etc.

cap10morgan: Good question. I have no idea.

I'll make it return something saner.

test/election_notification_works/handlers/vote_tomorrow_test.clj
```diff
@@ -65,9 +73,19 @@
...
+                   (util/user-district-division election user))))
+      (is (= #{:in-person :by-mail}
+             (-> ctx :district-division :voting-methods keys set))))))
+
+(deftest handle-update-test
+  (testing "The update handler adds in the :polling-place"
+    (with-redefs [vote-tomorrow/user-polling-place (constantly 42)]
+      (let [out (async/chan 1)
+            notification {}]
+        (is (= (-> (vote-tomorrow/handle-update out notification)
```
ericnormand: Please take from `out` directly.

test/election_notification_works/handlers/vote_tomorrow_test.clj
```diff
@@ -65,9 +73,30 @@
...
+              notification {:recipient {:addresses {:registered {:street "420 High St"
+                                                                 :city "Denver"
+                                                                 :state "CO"
+                                                                 :zip "80555"}}}}
+              result (-> (handle-update out notification)
+                         :notification
+                         :polling-place)]
+          (is (= fake-polling-place
+                 (dissoc result :map-url)))
+          (is (= "https://maps.google.com/maps?q=420+High+St+Denver+CO+80555+to+123+Main+St+Townsville+CO+87654"
```
ericnormand: Could you also test that the message gets put on `out`?

cap10morgan: Yep, good idea.

ericnormand:
LGTM

PR #33, opened 2015-12-15
Add transactional message sending.

ericnormand:
Yes, now you can send messages with a single POST.

However, there are two restrictions: 1, the user has to be subscribed to the medium and 2, the user's phone number has to match the recipient's phone number.

As of now only sms medium is accepted.

Done as part of [Pivotal Card](https://www.pivotaltracker.com/story/show/105595718)


dcohen:
LGTM.

PR #35, opened 2016-01-06
expect v2 elections embedded in electorates

cap10morgan:
While working on removing v1 elections, I realized something a bit subtle about electorates. While they embed elections in them, the electorate messages themselves are electorates, not elections. So unless we want to have version keys embedded all over the place, I think we should say that electorates have their own, independent version numbers and that's it. That's implicitly what [this electorate-works PR]() does, but I didn't realize the implications for this downstream component until now.

And really, we have some more details to figure out w/r/t versioning before it's a simple, reliable mechanism for uncoordinated deploys of breaking changes.

So what I'm proposing here is a companion PR to the aforementioned one in electorate-works that will just need to be deployed right around the same time as that one. With this PR, the one in electorate-works, and [this krakenstein PR](https://github.com/democracyworks/krakenstein/pull/10), krakenstein still sends election notifications.

I think we need more research on versioning in messages, but I'm not sure we'll get to it super soon.


ericnormand:
This looks good, but it's getting confusing.

Is this a regression or moving forward? I can't remember what the current version of the key is (:voting-methods or :voting-options).

cap10morgan:
Moving forward. voting-methods is the latest and greatest.

ericnormand:
Given what you've discovered, I agree versioning needs more work. I also think that we should worry about it later and just do manual synchronized deploys for now, so we can continue making progress.

LGTM

PR #37, opened 2016-01-27
attach VBM appl forms to signup emails

cap10morgan:
A part of [this Pivotal card](https://www.pivotaltracker.com/story/show/110044076)


src/election_notification_works/handlers/turbovote_signup.clj
```diff
@@ -24,16 +33,15 @@
...
             :medium :email ;; That's all we support
-            :context (merge recipient signers voter-reg
-                            (select-keys message [:voter-reg]))
+            :context (merge recipient signers
+                            (select-keys message [:voter-reg
+                                                  :vbm-application
+                                                  :election]))
             :identifiers {:user-id (:id user)}}
-           (when attachments {:attachments attachments})
+           (when (seq attachments) {:attachments attachments})
```
sduckett: ...because form-attachments might give an empty vector, which is truthy?

src/election_notification_works/handlers/turbovote_signup.clj
```diff
@@ -24,16 +33,15 @@
                                :email (:email user)}}
         signers {:signers (apply format "%s, %s, and %s"
                                  (util/signer-names))}
-        voter-reg (:voter-reg message)
-        form-uri (:form-uri voter-reg)
-        attachments (when form-uri [{:content-type "application/pdf"
-                                     :url form-uri}])]
+        attachments (form-attachments message)]
     (merge {:notification-type :turbovote-signup-confirmation
             :medium :email ;; That's all we support
```
sduckett: What about `,,, :medium supported-mediums ,,,` rather than a comment?

sduckett:
LGTM - my questions are more curiosity than blocking.

src/election_notification_works/handlers/turbovote_signup.clj
cap10morgan: You got it. :)

src/election_notification_works/handlers/turbovote_signup.clj
cap10morgan: I tried it out in the code and didn't like it much better. The context is that we would normally pull the requested medium from the `message` arg, but in this case we only support `:email`. So *really* what we should be doing is checking that `:email` was the requested medium and returning an error response if it wasn't. Hmm... Maybe I'll try that on for size.

cap10morgan: Looks like we aren't yet sending a `:medium` key from turbovote-web. So while there is a TODO comment a little below this asking for this very change, it's going to have to wait until after turbovote.org is pointed at the rewrite.

But I've done this side of it locally now, so we can probably mop this up on a Friday morning sometime.

PR #38, opened 2016-02-01
Upcoming election email ballot request

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/106113876)

The commit-by-commit view probably won't be useful here, it contains a couple large WIP commits.

So, what we're doing here is using the same scheduling update functionality the vote-tomorrow notifications use to add polling place info to render the actual notifications content. This helps us in two aspects: PDFs only hang around for 24 hours so we have to render them close to the time we send out the notifications, and we're more likely to have a user's voting-method-preference (`:by-mail` or `:in-person`) if we delay sending.

In addition, there have been upstream changes to put the user's matching district-division into the electorate, and we removed code that figured this out and started relying on the value coming in on the electorate. We also needed to update vote-tomorrow rendering because it was written under the assumption that there would only ever be a single voting-method for each type (again, `:by-mail` or `:in-person`) so we adjusted it to the new reality.

We have changes in [scheduled-message-works](https://github.com/democracyworks/scheduled-message-works/pull/13) and [notification-templates](https://github.com/democracyworks/notification-templates/pull/12) as well. Additionally, we'll be submitting a PR to krakenstein with our test scenarios.


tank157:
Assigning to @danielglauser

src/election_notification_works/handlers/util.clj
```diff
@@ -19,7 +19,7 @@
    "Seth"
    "Troy"
    "Wes"
-   "Wesley"])
+   "Sean"])
```
danielglauser: Nice

tank157: Arguably the most important change here.

tie-rack: Welcome, Sean. But also :sob: :sob: :sob:

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+  containing their preference, or nil if none is found."
+  [user-id lookup-fn]
+  (let [chan (lookup-fn {:user-id user-id})
+        response (async/alt!!
+                   chan ([x] x)
+                   (async/timeout 5000) {:status :timeout})]
+    (if (= :ok (:status response))
+      ;;TODO Bring the read response in line with normal mechanics
+      (condp = (:preference response)
+        :by-mail {:by-mail true}
```
danielglauser: Why do we need to explicitly set these values to `true`? Shouldn't whatever calls it be looking more for truthy/falsy instead of `true` and `false`?

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+        :in-person {:in-person true}
+        nil)
+      (log/error "Can't find a voting method preference" (pr-str response)))))
+
+(defn acceptable-form-for
```
danielglauser: I like this :+1:

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+           first
+           :name))
+
+(defn ->vbm-form
+  "Build a vbm form context for `recipient`, given their voting `preference`,
+  an `election`, and a `vbm-form-fn` that returns a channel. Returns nil
+  if there is no form to send."
+  [electorate preference vbm-form-fn]
+  (when-let [acceptable-form (acceptable-form-for electorate)]
+    (when (:by-mail preference)
```
danielglauser: Is there a case where we want to give a user a VBM form even if they have a different voting preference?

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+           (some :acceptable-forms)
+           (remove empty?)
+           first
+           :name))
+
+(defn ->vbm-form
+  "Build a vbm form context for `recipient`, given their voting `preference`,
+  an `election`, and a `vbm-form-fn` that returns a channel. Returns nil
+  if there is no form to send."
+  [electorate preference vbm-form-fn]
```
danielglauser: I like the pattern of passing `fn`s around instead of channels with answers.

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+  (when-let [acceptable-form (acceptable-form-for electorate)]
+    (when (:by-mail preference)
+      (let [user (:user electorate)
+            voter-info {:form acceptable-form
+                        :voter (select-keys user [:first-name :last-name])}
+            vbm-chan (vbm-form-fn voter-info)
+            response (async/alt!!
```
danielglauser: Do you think it's worth creating a `take-with-timeout` `fn`? This seems to repeat throughout this and other `ns`es. Since the code isn't very complex I could go either way here.

src/election_notification_works/handlers/upcoming_election_update.clj
```diff
@@ -0,0 +1,199 @@
...
+          (when-let [form-uri (:form-uri response)]
+            [{:content-type "application/pdf"
+              :url form-uri}])
+          (log/error "Hoping for a form, didn't get one" (pr-str response)))))))
+
+(defn handle
+  "Takes an incoming 'notification` with an `electorate`, and builds a
+  context.
+
+  ```{:recipient {:full-name ...
```
danielglauser: This is great for now, we may want to think about additional Schemas. I find them a double edged sword so I'm not pushing for that now, just something to think about.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -75,29 +76,21 @@
...
+(defn electorate->ctx
+  "Given an electorate, produce the context with description and voting
+   id instructions"
+  [electorate]
+  (let [description {:description (get-in electorate [:election :description])}
+        instructions (some->> electorate
+                              :district-division
+                              :voting-methods
+                              (filter vm/in-person?)
+                              (some #(get-in % [:instructions :voting-id]))
```
danielglauser: So we just grab the first set of instructions? Is product cool with this?

src/election_notification_works/voting_methods.clj
```diff
@@ -0,0 +1,32 @@
+(ns election-notification-works.voting-methods
```
danielglauser: This `ns` is great. I wonder if it will need to be broken out.

test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -120,24 +93,23 @@
                                       (coerce/to-date
                                        (time/plus (time/now)
                                                   (time/days 23))))]
-      (handle results electorate #{:sms :email})
+      (handle results electorate #{:sms :email} )
```
danielglauser: whitespace

test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -153,51 +125,9 @@
            first-message (read-response results-chan)
            second-message (read-response results-chan)
            email-message (first (filter #(= :email (:medium %)) [first-message second-message]))
-           sms-message (first (filter #(= :sms (:medium %)) [first-message second-message]))]
+           sms-message (first (filter #(= :sms (:medium %)) [first-message second-message]))
+           ]
```
danielglauser: You left a dangler

test/election_notification_works/handlers/upcoming_election_update_test.clj
```diff
@@ -0,0 +1,189 @@
...
+(defn vbm-form-fn [_] (async/go {:status :ok :form-uri :a-uri}))
+
+(deftest same-day-voter-reg-test
+  (testing "when does an election allow same day voter registration?"
+    (let [electorate (test-electorate)]
+      (is (true? (same-day-voter-reg? electorate))))))
+
+(deftest ->recipient-test
+  (testing "uses :user/nickname for recipient.first-name when present"
+    (let [nickname "Stinky"
```
danielglauser: http://vignette2.wikia.nocookie.net/clubpenguin/images/d/d6/Stinky_Cheese_C-J_Card_728.png/revision/latest?cb=20130608095546

test/election_notification_works/handlers/upcoming_election_update_test.clj
```diff
@@ -0,0 +1,189 @@
...
+
+(use-fixtures :each with-fresh-db)
+
+(defn lookup-fn [_] (async/go {:status :ok}))
+(defn vbm-form-fn [_] (async/go {:status :ok :form-uri :a-uri}))
+
+(deftest same-day-voter-reg-test
+  (testing "when does an election allow same day voter registration?"
+    (let [electorate (test-electorate)]
+      (is (true? (same-day-voter-reg? electorate))))))
```
danielglauser: truthy/falsy vs. testing for discrete values?

src/election_notification_works/handlers/upcoming_election_update.clj
tank157: We do this because Handlebars sucks. When doing a {{#if something.something.by-mail}} type of conditional, you need a value AT the key, not just the key itself.

src/election_notification_works/handlers/upcoming_election_update.clj
tank157: There may be that in the future, but for now, that wasn't in the template's requirements.

src/election_notification_works/handlers/upcoming_election_update.clj
tank157: Hat Tip to Eric for that, I think. Though, the `fn` still returns a channel with an answer, but that's dead simple to simulate in testing with `async/go`.

src/election_notification_works/handlers/upcoming_election_update.clj
tank157: Yeah, it's another function-let that shows up lots of places, and we don't have a good story on where something like that should exist. I don't mind it being repeated, it tends to be small, and the syntax for `alt!!` is confusing enough with the `([x]...)` structure that I somewhat like seeing it around and re-inforcing that odd mechanic, but that's just me.

test/election_notification_works/test_helpers.clj
```diff
@@ -22,6 +22,8 @@
    (async/alt!! chan ([v] v)
                 (async/timeout timeout) :timeout)))
 
+(defn make-user-id [] (java.util.UUID/randomUUID))
```
danielglauser: I wonder if we could standardize on a name here, like `random-uuid`?

src/election_notification_works/handlers/vote_tomorrow.clj
tank157: We're grabbing the first instructions from any `:in-person` voting methods only. Right now, there's only ever one. At some point in the future we'll have a second one, for `:early-voting`, but we don't have the data model figured out to distinguish early voting from voting day, so it's hard to figure out just what to distinguish with here at the moment.

src/election_notification_works/handlers/upcoming_election_update.clj
danielglauser: That's legit.

src/election_notification_works/handlers/upcoming_election_update.clj
danielglauser: Okay

danielglauser:
@tank157 @sduckett Done dropping comments.

test/election_notification_works/test_helpers.clj
tank157: Good point, nothing about it really says that it's a user-id, will change.

test/election_notification_works/handlers/upcoming_election_update_test.clj
tank157: This is code that migrated wholesale from the old `ns` to the new one, so we didn't exactly "change" it, but happy to see if we can make it a little better.

test/election_notification_works/handlers/upcoming_election_update_test.clj
tank157: I thought you were pointing out a bad code smell for a bit there.

src/election_notification_works/handlers/vote_tomorrow.clj
danielglauser: Fair enough

danielglauser:
Well done sirs. LGTM

danielglauser:
LGTM

PR #39, opened 2016-02-09
Fix upcoming election already past

tank157:
Should fix [this bug](https://www.pivotaltracker.com/story/show/113007779) and close out [this card](https://www.pivotaltracker.com/story/show/106113876). The solution involves scheduling notifications that would go out right away for 10 minutes into the future, so that we can still get the most complete voter record we can, including voting-method-preferences.

One last review requested by @danielglauser "The Ocho"


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -19,68 +236,68 @@
...
-         coerce/from-date)))
+  [medium tz send-local-date]
+  (let [{:keys [hour minute]} (medium->time medium)]
+    (-> (coerce/to-date-time send-local-date)
+        (time/from-time-zone tz)
+        (time/plus (time/hours hour))
+        (time/plus (time/minutes minute)))))
 
 (defn ideal-send-at
   ([medium electorate] (ideal-send-at medium electorate util/NY))
```
danielglauser: Please add a doc string that describes what this returns.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -1,9 +1,226 @@
...
+    (if (= :sms medium)
+      {:description election-description
+       :date (->> election-date
+                  (fmt/unparse sms-date-formatter))}
+      {:description election-description
+       :date (->> election-date
+                  (fmt/unparse email-date-formatter))})))
+
+(defn ->deadlines
+  "Calculate the the deadlines for voter registration and voting by mail
```
danielglauser: the the

danielglauser:
Once "the" typo is fixed LGTM from @7maples and I.

PR #40, opened 2016-02-17
Check responses

cap10morgan:
Previously we were ignoring responses to our message scheduling in election-notification-works. That's bad, we should check them instead and do something useful if they are errors.

This is the 2nd half of [this card](https://www.pivotaltracker.com/story/show/111073346).


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -286,22 +291,23 @@
...
-                         mediums)]
-      (when notifications
-        (log/debug "Scheduling upcoming-election notification(s):"
-                   (pr-str notifications))
-        (util/send-notifications schedule-channel notifications)))))
+    (let [notifications (->upcoming-notification-messages electorate mediums)]
+      (log/debug "Scheduling upcoming-election notification(s):"
+                 (pr-str notifications))
+      (doseq [n notifications]
+        (let [response (async/<!! (schedule-fn n))]
```
danielglauser: Should we have a timeout here?

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -138,30 +138,36 @@
...
+  sends any notifications with the given `send-fn`."
+  [send-fn electorate mediums]
   (log/debug "vote-tomorrow handling an electorate:" (pr-str electorate))
   (let [election (:election electorate)]
     (when-not (too-late-for-notification? election)
       (let [notifications (->vote-tomorrow-notification-messages mediums
                                                                  electorate)]
-        (util/send-notifications channel notifications)))))
+        (doseq [n notifications]
+          (let [response (async/<!! (send-fn n))]
```
danielglauser: Timeout as well?

src/election_notification_works/handlers/upcoming_election.clj
cap10morgan: No need. [Kehaar closes the response channel if it hits its own timeout](https://github.com/democracyworks/kehaar/blob/master/src/kehaar/wire_up.clj#L112-L114).

cap10morgan: Though I *should* handle the nil value that will be returned in that situation. :)

danielglauser:
@cap10morgan I think this is better but won't we still hit an issue with the blocking takes if nothing ever ends up on the channel? I was thinking something like this: https://github.com/democracyworks/address-works/blob/master/test/address_works/queue_message_test.clj#L18-L19

cap10morgan:
@danielglauser No because kehaar is already handling that for these channels. If nothing shows up by the time its timeout hits, it will close the channel, which will cause the blocking take to immediately return `nil`.

danielglauser:
@cap10morgan Nice! Looks like I totally missed that in Kehaar. LGTM

PR #41, opened 2016-02-23
Respond to electorate.delete events

ericnormand:
Electorate-works is now sending electorate.delete events when an
electorate is removed (in response to the delete command or based on a
user changing addresses). These cascade into multiple
notification.unschedule events, based on medium and type of
notification.

[Pivotal Card](https://www.pivotaltracker.com/story/show/113401679)


ericnormand:
Note: I'm going to run a final krakenstein test to make sure this is all working since i made some cosmetic changes. I'll report the results here.

ericnormand:
Update: krakenstein tests are positive!

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -305,3 +306,19 @@
...
+(defn handle-delete
+  "Takes an incoming `electorate` and `mediums` and unschedules them."
+  [unschedule-channel electorate mediums]
+  (let [notifications (->upcoming-notification-messages
+                       electorate
+                       mediums)]
+    (when (seq notifications)
+      (log/debug "Unscheduling upcoming-election notification(s):"
+                 (pr-str notifications))
+      (util/send-notifications
```
cap10morgan: Since all `util/send-notifications` does is call `async/onto-chan` it seems better to just call that directly here instead of a fn whose name implies that it is strictly for sending notifications (which this is not doing). If we made changes in the future to `send-notifications` under the assumption that it is all it was used for, it could break this.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -165,3 +165,21 @@
...
+(defn handle-delete
+  "Takes an incoming `electorate` and mediums,
+  checks for valid date range, and sends any
+  notifications needed onto the provided
+  sending-fn."
+  [channel electorate mediums]
+  (let [notifications (->vote-tomorrow-notification-messages mediums
+                                                             electorate)]
+    (when (seq notifications)
+      (log/debug "vote-tomorrow unscheduling a notifications:"
```
cap10morgan: Singularize `notifications`

src/election_notification_works/queue.clj
```diff
@@ -118,6 +127,11 @@
                               ""
                               "notification-works.notification.schedule"
                               channels/notification-schedule)
+                             (wire-up/outgoing-events-channel
```
cap10morgan: This should be an `external-service` now, and we should check the responses and do something appropriate when they are errors.

src/election_notification_works/handlers.clj
```diff
@@ -116,7 +116,11 @@
 
 (def notification-handlers
   #{(partial upcoming-election/handle channels/notification-schedule)
-    (partial vote-tomorrow/handle channels/notification-schedule)})
+    (partial     vote-tomorrow/handle channels/notification-schedule)})
```
sduckett: stray whitespace?

src/election_notification_works/handlers/vote_tomorrow.clj
sduckett: `handle-delete` doesn't seem to have tests, unless they're outside of this project somewhere. Did I miss something?

src/election_notification_works/handlers/upcoming_election.clj
ericnormand: This will get cleaned up when I change it to services with responses.

src/election_notification_works/handlers.clj
ericnormand: Just lining them up.

ericnormand:
Alrighty, I addressed everything here. Any new blockers?

cap10morgan:
I have another PR open here that checks responses from notification-works. We should merge that and then make this conform to what it's doing (unless you don't like anything that it's doing) before we merge this. But I'd be happy to do that once my PR is merged. It's close; I don't think it will be a big change.

cap10morgan:
LGTM otherwise.

cap10morgan:
That other PR is merged now, @ericnormand. Let me know if you'd rather do the merge with that. Otherwise I'm happy to take a crack at it and let you review the result.

ericnormand:
Just merged from master. I added error handling like you had. I'm testing again with krakenstein.

cap10morgan:
Awesome

ericnormand:
Ok, works with krakenstein's monster.

sduckett:
LGTM

PR #44, opened 2016-03-21
[TH] remove hard coded 5 second timeout, use config timeout

tank157:
Fixing a production timeout issue


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -139,8 +140,11 @@
                         :voter (select-keys user [:first-name :last-name])}
             vbm-chan (vbm-form-fn voter-info)
             response (async/alt!!
-                       (async/timeout 5000) {:status :timeout}
-                       vbm-chan ([x] x))]
+                       (async/timeout (config [:voter-apply-timeout]))
+                       {:status :timeout}
+
+                       vbm-chan
+                       ([x] x))]
```
cap10morgan: Shouldn't we just get rid of this `alt!!` altogether? It's redundant with kehaar's own timeout.

tank157: All in due time, just trying to fix the bug before it generates a mountain of manual fix work, and this is the quickest path to that.

sduckett:
LGTM.

PR #47, opened 2016-05-18
Update signers list for email notifications

marlabrizel:
Update names on signers list
Replace ‘and’ with ‘&’ in signature text


7maples:
LGTM

PR #48, opened 2016-05-24
Verify notifications

cap10morgan:
This is the code Marla and I used to find and re-schedule the missing notifications. Hopefully we won't need it again, but just in case we do...

[Pivotal card](https://www.pivotaltracker.com/story/show/119913437)


dev-src/election_notification_works/send_notifications.clj
```diff
@@ -0,0 +1,43 @@
+(ns election-notification-works.send-notifications
+  (:require [datomic.api :as d]
+            [election-notification-works.handlers.vote-tomorrow :as hvt]
+            [election-notification-works.handlers.upcoming-election :as hue]
+            [election-notification-works.handlers :as handlers]
+            [kehaar.wire-up :as wire-up]
+            [clojure.core.async :as async]))
+
+(def prod-rabbit-config {:host "10.36.116.190" :port 5672})
```
danielglauser: Did you mean to include the production IP?

dev-src/election_notification_works/send_notifications.clj
```diff
@@ -0,0 +1,43 @@
...
+   ""
+   "notification-works.notification.schedule"
+   {:exclusive false :durable true :auto-delete false}
+   10000
+   notification-schedule-chan))
+
+(def schedule-notification (wire-up/async->fn notification-schedule-chan))
+
+;; Useful for when you only want to send emails, not SMS's
+#_(defn email-subscription? [user]
```
danielglauser: Why is this commented out?

dev-src/election_notification_works/send_notifications.clj
cap10morgan: Yeah, it's not reachable on the public Internet. But it's also not critical to have in here (it's easy enough to specify manually in the REPL).

dev-src/election_notification_works/send_notifications.clj
```diff
@@ -0,0 +1,43 @@
...
+    (contains? (set (handlers/filter-subscription-mediums)
+                    (handlers/user-subscription user) user)
+               :email))
+
+(defn vote-tomorrows [electorate]
```
danielglauser: Maybe rename this to `schedule-vote-tomorrows`?

dev-src/election_notification_works/send_notifications.clj
```diff
@@ -0,0 +1,43 @@
...
+  (let [user (:user electorate)
+        mediums (handlers/filter-subscription-mediums
+                 (handlers/user-subscription user) user)]
+    (hvt/handle schedule-notification electorate mediums)))
+
+(defn upcoming-elections [electorate]
```
danielglauser: Maybe rename this to `schedule-upcoming-elections`?

dev-src/election_notification_works/send_notifications.clj
cap10morgan: Because it's only useful in specific circumstances (as the comment says), not the more general case. We wrote it to help with getting emails only out for some users, and it seemed a shame to just delete it in case we needed to do that again.

dev-src/election_notification_works/send_notifications.clj
cap10morgan: The namespace is intended to bring that bit of context to it.

dev-src/election_notification_works/verify.clj
```diff
@@ -34,27 +42,60 @@
...
+(defn verify-election-notifications
+  [enw-db smw-db electorates notification-type]
+  (let [counts (reduce (fn [sums electorate]
+                         (let [subscription (:mediums (subscription/read-subscription
+                                                        enw-db (get-in electorate [:user :id])))
+                               scheduled-noties (find-scheduled-noty notification-type
+                                                                     smw-db electorate)
+                               scheduled-notys-count (count scheduled-noties)
+                               subscription-count (count subscription)]
+                           (as-> sums $
```
danielglauser: Nice!

dev-src/election_notification_works/send_notifications.clj
danielglauser: Got it, sounds good.

dev-src/election_notification_works/send_notifications.clj
danielglauser: Answered above.

dev-src/election_notification_works/send_notifications.clj
danielglauser: Hmmmm...the name of the function is such that I would never suspect that it would apply to SMS messages. I'd prefer to uncomment it, this code does live in `dev-src` which already means it's outside of our normal workflow (yes, I consider imports and exports outside of our normal workflow).

cap10morgan: OK, I don't have a strong opinion about it. I'll uncomment it.

dev-src/election_notification_works/verify.clj
cap10morgan: @tank157 suggested that little nugget.

dev-src/election_notification_works/verify.clj
```diff
@@ -64,14 +105,31 @@
...
-        (println "Finding electorates for" (:description election) "-" (:id election))
-        (let [electorates (async/<!! (electorate-search {:election-id (:id election)}))]
-          (verify-upcoming-election-notifications scheduled-message-works-db
-                                                  electorates))
-        #_(verify-vote-tomorrow-notifications electorate)))
+        (println "Finding electorates for" (:description election) "on" (:date election) "-" (:id election))
+        (let [{:keys [electorates]} (async/<!!
+                                     (electorate-search
+                                      {:election-id (:id election)}))]
+          (verify-election-notifications election-notification-works-db
```
danielglauser: I *think* I'm okay with this here since it's in a `dev-src` `ns`, however I think connecting to the data store of another microservice is an antipattern. I believe this should drop messages on queues. What do you think?

cap10morgan: Yeah ideally, but I don't think we should spend any time doing that right now.

danielglauser: Agreed.

danielglauser:
LGTM

PR #49, opened 2016-05-26
only add language parameter if it was present

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/120366137)

Once this work is deployed, I have 9 messages to resend manually (will test one to us first).

Assigning to @danielglauser


danielglauser:
LGTO <--- Looks Good To Ocho

PR #50, opened 2016-06-07
Add election-mail subscription status to template contexts

cap10morgan:
...so we can use it to change what we tell users in some notifications.

[Pivotal card](https://www.pivotaltracker.com/story/show/119219199)

@ddellacosta and I are in the process of testing this locally.

There is some refactoring of election-notification-works where we found problems (like the invalid context when scheduling upcoming-election notifications–which was being corrected in the update pass) and/or felt that adding in the election-mail subscription status would make hard-to-follow code even harder to follow. But let one of us know if you'd like a walkthrough of our thought process anywhere.


src/election_notification_works/handlers.clj
```diff
@@ -64,15 +55,15 @@
       (let [updated-subscription (subscription/update-subscription
                                   subscription-entity update)]
         (when-not (:suppress-confirmation update)
-          (confirmation/handle send-notification
+          (confirmation/handle notification/send
```
sduckett: I like using a `notification` namespace here

src/election_notification_works/handlers.clj
```diff
@@ -122,8 +113,8 @@
           (:mediums subscription)))
 
 (def notification-handlers
-  #{(partial upcoming-election/handle schedule-notification)
-    (partial vote-tomorrow/handle schedule-notification)})
+  #{upcoming-election/handle
```
sduckett: I'm trying to remember why partial applications were used here, but it seems to read a little better without.

src/election_notification_works/handlers/sms_form.clj
```diff
@@ -27,14 +29,15 @@
...
+(defn handle
+  [message]
+  (s/validate send-schema message)
   (let [sub (subscription/read-subscription (db/db) (:user-id message))]
     (assert-or-throw (contains? (:mediums sub) (:medium message))
                      "The requested medium is not subscribed to by the user."
                      message)
-    (let [{:keys [status user]} (async/<!! (read-user-fn {:id (:user-id message)}))]
+    (let [{:keys [status user]} (async/<!!
+                                 (read-user-fn {:id (:user-id message)}))]
```
sduckett: Maybe `read-user-fn` could be `read-user`? I'm not convinced that this naming style is beneficial.

src/election_notification_works/handlers.clj
cap10morgan: IIRC it was so we could inject stubbed `schedule-notification` fns in tests. But there's already a (IMO) better approach there that uses the real scheduling fn's and just simulates kehaar on the other ends of the real core.async channels. I think it's better because–among other reasons–it runs through more of the production code in the tests. We've switched to that approach in here.

src/election_notification_works/handlers/subscription_confirmation.clj
```diff
@@ -12,10 +12,7 @@
 
 (defn ->confirmation-notification-messages
   [user mediums]
-  (let [recipient {:recipient {:full-name (util/full-name user)
-                               :first-name (util/preferred-first-name user)
-                               :email (:email user)
-                               :phone (:phone user)}}
+  (let [recipient {:recipient (util/user->recipient user)}
         signers {:signers (apply format "%s, %s, & %s"
```
sduckett: @ddellacosta's not in `util/signer-names` yet.

cap10morgan: Ack! Will fix! :)

test/election_notification_works/test_helpers.clj
```diff
@@ -17,13 +17,27 @@
...
-(defn read-response
+(defn take-or-timeout
   ([chan]
-   (read-response chan 500))
+   (take-or-timeout chan 500))
   ([chan timeout]
    (async/alt!! chan ([v] v)
                 (async/timeout timeout) :timeout)))
 
+(defn stub-external-service
```
sduckett: This feels like a style that will eventually be found elsewhere. I wonder: is there a common set of async-related test helpers scattered across the `-works` set?

cap10morgan: Yeah, pretty much. We should create a library at some point. Or maybe even add a `test-helpers` namespace to kehaar.

test/election_notification_works/handlers/vote_tomorrow_test.clj
```diff
@@ -5,9 +5,10 @@
             [clojure.core.async :as async]
             [clj-time.core :as time]
             [clj-time.coerce :as coerce]
-            [election-notification-works.test-helpers :refer :all]
+            [election-notification-works.test-helpers :refer :all :as helpers]
```
sduckett: What's going on with `:refer :all :as helpers`? I'd expect to see `:refer :all` _or_ `:as helpers` but not a combination.

cap10morgan: We typically `:refer :all` the test-helpers but I also needed to refer to a namespaced keyword on line 118 and this allowed me to write `::helpers/timeout` there instead of `:election-notification-works.test-helpers/timeout`.

sduckett:
I see templates changing in [another PR](https://github.com/democracyworks/notification-templates/pull/20), so it seems like this covers the ground of the referenced Pivotal card. There's a lot to read in this PR, but it LGTM.

cap10morgan:
@sduckett 2 new commits to review please?

sduckett:
@cap10morgan I like things that make testing in a REPL _really simple_. I'd be keen on a better name for `assoc-paper-vbm-form-status` (I don't have one....yet...), but functionally it LGTM.

PR #52, opened 2016-06-21
Fix election-mail-subscription in sms-form more

cap10morgan:
Last "fix" inadvertently added a new top-level :recipient key rather
than adding the election-mail-subscription to the existing one under the
:context map.


danielglauser:
LGTM

PR #55, opened 2016-07-06
Upcoming election update fixes

tank157:
This fixes the upcoming election update scheduled message workflow.

[Pivotal Card](https://www.pivotaltracker.com/story/show/124816691)

Assigning to @cap10morgan, also @danielglauser should take a look.

I have run this code in krakenstein, generated the scheduled messages, and then sent them so they hit the update handler as well, and got the expected results. I only used one test scenario, but it does do 3 different voters, so reasonably confident it works.

Any currently scheduled message won't work, but we're planning to rerun them all.


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -168,21 +178,10 @@
       :same-day-voter-reg true
       :voting-method {:all-mail true} ;;or :excuse-required, :no-excuse-vbm
       :deadlines {:vbm {:date \"Formatted Date\"}}
-      :signers \"Matt, Matt and Matt\"}```
-
-  In addition, should the voter's preference by to vote by mail and there
```
cap10morgan: s/preference by/preference be/?

test/election_notification_works/test_helpers.clj
```diff
@@ -10,7 +10,7 @@
 (defn random-uuid [] (java.util.UUID/randomUUID))
 
 (defn with-fresh-db [test]
-  (let [uri (str (config [:datomic :uri]) "-" (random-uuid))]
+  (let [uri (str "datomic:mem://election-mail-works-" (random-uuid))]
```
cap10morgan: `election-mail-works`??

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -203,22 +201,77 @@
...
+                      :url \"url of rendered form\"}]```"
+  [notification]
+  (let [medium (:medium notification)
+        user-id (get-in notification [:context :recipient :id])
+        election-id (get-in notification [:context :election :id])
+        electorate (get-electorate user-id election-id)]
+    (if electorate
+      (let [voting-preference (->voting-method-preference user-id)
+            election-mail-subscription (election-mail/user-subscribed?
+                                        (:user electorate))
```
cap10morgan: We really, really need to stop doing this in two different ways (update loopbacks and event handlers). We're building the election-mail-subscription piece according to our architecture (i.e. with event handlers) as I type this. So that should get taken out of here once that work is done.

We should figure out a secure form storage solution and add event handling to voting method preferences too and then stop doing the loopback for anything that we can know ahead of time or get events for when it is updated.

But in the meantime, we're just going to have to live with this and be careful.

test/election_notification_works/test_helpers.clj
tank157: Good catch, I think I was looking at it at the time I changed this and brain didn't switch for me.

src/election_notification_works/handlers/upcoming_election.clj
tank157: Definitely agree with you. This is an expedient fix right now, to stem the tide of bleeding. I am hesitant to add new event driven things until we know what's going on with the missing electorates isn't anything to do with event processing itself. But yes, this should be handled with triggered updates based on voting-method-preference and election-mail-subscription events.

danielglauser:
I glad you tested this the way you did and I think this should work. My main question is shouldn't it be an error if we don't find the electorate when updating the notification?

tank157:
Not necessarily. I don't think we have all the plumbing linked up to properly clean things when we remove either an election or a user or an electorate. Even if we did, there's still always the possibility of someone changing their information concurrent with the scheduled message firing, and the process that would have deleted the scheduled message hasn't arrived yet. It will hopefully just be a rare possibility though.

danielglauser:
LGTM

PR #56, opened 2016-07-11
Zero was not a valid value for :send-after

danielglauser:


tank157:
Thumbs up once it passes build checks

PR #58, opened 2016-07-20
User delete event

tank157:
Handles cleanup of subscriptions when a `user.delete` event comes in.

I also refactored a bit of work from the previous PR, to stop passing channels into the `create-subscription` and `delete-subscription` handlers. Dave and I still have thoughts about this, there are compromises both ways, but I do like getting out of a level of partialization.

Will be testing out in krakenstein shortly with [another PR](https://github.com/democracyworks/electorate-works/pull/38) for `electorate-works`.

Not strictly tied to a feature card, a lot of this was done after hours, but tangentially related to making the chore cards for deleting accounts simpler. With this, we won't have to manually go find the subscriptions and delete them as well.


tank157:
Assigning to @7maples

src/election_notification_works/handlers.clj
```diff
@@ -200,3 +200,11 @@
          response
          (str "Error unscheduling notification "
               (pr-str message)))))))
+
+(defn handle-user-delete
+  "When a user is deleted, find and deletes subscriptions, which in
+   turn can unscheduled future notifications"
```
7maples: small typo, should be "unschedule"

7maples:
LGTM

PR #59, opened 2016-07-21
Mail subscription events

cap10morgan:
This handles election-mail subscription events (create & delete) by updating scheduled notifications whose content depends on them. It solves a race condition on signup where a user's notifications would get scheduled w/ `:election-mail-subscription false` before they were asked whether they wanted paper mailings or not. Now when/if they opt in, that event will cause their notifications to get updated.

It was kind of a pain, and should inform a significant refactor of the notifications subsystem next year.

But it does work in krakenstein.

Note that this switches to using `:recipient-id` for turbovote-signup notifications, which is part of [this bug card](https://www.pivotaltracker.com/story/show/126546229). After this is deployed, we should manually update `:user-id` identifiers in the production DB to be `:recipient-id` instead in scheduled-message-works.

Once merged we should stop updating this in places [like this](https://github.com/democracyworks/election-notification-works/blob/master/src/election_notification_works/handlers/upcoming_election.clj#L259-L261) and anything similar in other handlers (as it'll no longer be necessary).

This is dependent on other PRs in [election-mail-works](https://github.com/democracyworks/election-mail-works/pull/12) and [notification-works](https://github.com/democracyworks/notification-works/pull/28). We should discuss deployment coordination once they all have the LGTM. It _should_ be pretty straightforward.
## Pivotal cards this addresses
- [Original feature card which is already accepted but shouldn't have been](https://www.pivotaltracker.com/story/show/119219199)
- [First bug card which was written before the feature was done for some reason](https://www.pivotaltracker.com/story/show/121900847)
- [Second bug card which was also written before the feature was done](https://www.pivotaltracker.com/story/show/121900335)
- [Third bug card which was, yes, written around the same time as the first two](https://www.pivotaltracker.com/story/show/121950045)


cap10morgan:
@tank157 Upon looking it over again, I'd really like to keep the `:user-id` -> `:recipient-id` fix in here.  It feels like the kind of thing that will be easy to forget about and/or de-prioritize into oblivion only for it to bite us in the future. Since I'm planning on moving scheduled-message-works over to The Death Star anyway, how about this:

- After shutting down scheduled-message-works in the old cluster but before starting it up in the new one...
    - I'll deploy this to the new cluster.
    - I'll run a Datomic transaction to change every `:user-id` identifier in the DB to `:recipient-id`.

Then I'll fire up s-m-w in the new cluster.

tank157:
I'm a bit nervous about the amount of downtime it still might take. For one, we'll be getting timeouts unless we have fire-and-forget deployed in `election-mail-works` and `election-notification-works` and possibly also `ballot-notification-works`. Also, changing this value might take some time to accomplish. We're talking a significant portion, if not almost all of the scheduled notifications probably have one, and so that's a lot of data to find/update/transact and then also a lot of indexes will have to be adjusted by the transactor. It's simple in concept, but possibly potentially thorny in execution.

cap10morgan:
And yet, I'd still much rather get it fixed. I'm not worried about innocuous timeouts. Users won't be affected by them.

Indexes are updated asynchronously, so that won't slow down the data correctness.

I think we can do this without causing a lot of problems. But I'd like to do it tomorrow if possible.

tank157:
OK, well, I don't see the risk as worth the value, but if you want to take on the duties of the migration and any fallout, I guess go for it. I'd recommend finding a quiet time to do it, since the identifiers have been a delicate complication lately, and if the indexes are rebuilding and it has to resort to some kind of linear scan it will probably hurt performance for a while, though mostly just on deletes.

tank157:
@cap10morgan I've remerged master again, and made a number of changes to get this working. Would you like to take another look? I managed to deal with the comment about "there should be a better way to search multiple notification types in `sms-form` handler, cause, well, there was. And a number of other changes to just make things work and to work properly. There are also changes in `scheduled-message-works` and `election-mail-works` to get this all working.

cap10morgan:
LGTM

PR #60, opened 2016-08-02
expect full user on user.delete event

tank157:
This is the only other place (aside from electorate-works) currently listening to `user.delete` events, and so needed to be updated to expect the full user, not just the id.

Assigning to @ddellacosta


ddellacosta:
LGTM

PR #61, opened 2016-09-02
Cleanup turbovote signup

danielglauser:
This will make it easier to retry failed `turbovote-signup` requests and will show us where `turbovote-web` isn't sending `:delay-minutes` because that request will now fail a schema validation. In support of [this](https://www.pivotaltracker.com/story/show/129728469) ticket.


tank157:
LGTM

PR #62, opened 2016-09-05
added "and the rest of the TurboVote team" after signers

kep:
Changed based on recommendation from user testing; without context, users do not understand who the three names belong to.


cap10morgan:
Yes! I have long wanted to see this brought over from classic.

Nittiest of nitpicks: Does it need to be in parentheses? "Matt, Dave, Kathryn, and the rest of the TurboVote team" seems more straightforward and would just be:

```clojure
(apply format "%s, %s, %s, and the rest of the TurboVote team"
                                   (util/signer-names))
```

kep:
SGTM, if you want to make that edit. If you do, can we keep the & for
"and"? I like the curlicue.

cap10morgan:
Yeah, I'll keep the &. :)

cap10morgan:
OK, that's pushed. Along w/ a centralization of the code that generates that string (so we only have to change it in once place in the future) and an updating of the list of signer names that the system chooses 3 random elements from.

ddellacosta:
LGTM

PR #69, opened 2016-09-10
Take medium into account when deciding between create and update

danielglauser:
Thought not as good as using `swap-tx!` this should take care of the common race condition we were seeing.


danielglauser:
[Pivotal card](https://www.pivotaltracker.com/story/show/121901547)

PR #70, opened 2016-09-11
Medium race condition swap tx

danielglauser:


src/election_notification_works/handlers.clj
```diff
@@ -60,40 +60,25 @@
...
-        subscription (if existing
-                       (update-subscription existing message)
-                       (create-new-subscription message))]
-    (when (not (= subscription
-                  (subscription/entity->subscription existing)))
-      (async/put! channels/subscription-create-events-out subscription))
+  (let [{:keys [db-before db-after]} (subscription/upsert-subscription message)
+        before (subscription/read-subscription db-before user-id)
+        after  (subscription/read-subscription db-after user-id)
+        new-mediums (set/difference (:mediums after) (:mediums before))]
```
ddellacosta: I don't love all the data wrangling above being in handler code, seems conceptually like it should be in the province of the entity upsert code--seems reasonable for an upsert operation to return the current value of the entity after updating.

tank157: Well, then the other option is to push channel operations into the entities package, which is arguably a lot worse.

ddellacosta: I'm just talking about getting the mediums as updated by the db, basically the four lines of code above my comment. That wouldn't require any channel futzing-about in the entity namespace would it?

ddellacosta:
Thanks @tank157 for explaining how `:assert-empty` works in Slack, very helpful. I still think the specific lines I was talking about in the `upsert-subscription` handler should be inside the entity namespace somehow, but I'll leave it up to you how far you want to go with tweaking that--I don't think it's worth holding up the PR over. LGTM!

src/election_notification_works/handlers.clj
danielglauser: @ddellacosta The four lines of code above your comment figure out what the `fn` in the entity `ns` did and use that information to decide if the handler should send a notification and if it should broadcast a change. The entity code does all the updating and manages the changes, the handler takes care of next steps based upon what was done.

PR #71, opened 2016-09-18
Trace all the things with NewRelic

danielglauser:


src/election_notification_works/deadlines.clj
```diff
@@ -1,8 +1,9 @@
 (ns election-notification-works.deadlines
   (:require [clj-time.coerce :as coerce]
-            [election-notification-works.handlers.util :as util]))
+            [election-notification-works.handlers.util :as util]
+            [com.climate.newrelic.trace :refer [defn-traced]]))
 
-(defn voter-reg-by-mail-deadline
+(defn-traced voter-reg-by-mail-deadline
```
tank157: I'm not really a fan of indiscriminately tracing every function, but actually looking at what the function does and making some decision based on that. When it's making a calculation based on local data as this does, tracing seems mostly like extra noise. Not to say there isn't the possibility of an algorithmically intensive hotspot that could be dealt with, but I just can't see that for code like this.

In general I'd add tracing at the edges of the systems, so handler, external call points (whether they're our own or third party systems), probably at datomic input/output points. But I really don't think just every function should be traced.

src/election_notification_works/election_mail.clj
```diff
@@ -2,12 +2,13 @@
...
             [clojure.core.async :as async]
-            [clojure.tools.logging :as log]))
+            [clojure.tools.logging :as log]
+            [com.climate.newrelic.trace :refer [defn-traced]]))
 
 (def read-subscription
   (wire-up/async->fn channels/election-mail-subscription-read))
 
-(defn user-subscribed?
+(defn-traced user-subscribed?
```
tank157: Like here, I can get behind this tracing because it can illustrate if making the system call to find out if a user has an election-mail subscription is intensive (and which it might be a bit right now as we don't have the index merged there).

src/election_notification_works/entities/subscription.clj
```diff
@@ -3,34 +3,35 @@
...
+            [com.climate.newrelic.trace :refer [defn-traced]]))
 
 (tx/deftx-data-fn subscription->tx-data
   [user-id mediums]
   {:db/id ?id
    :user/id user-id
    :notification/medium mediums})
 
-(defn entity->subscription
+(defn-traced entity->subscription
```
tank157: But tracing this is just overkill.

src/election_notification_works/handlers.clj
```diff
@@ -100,21 +101,21 @@
...
-(defn create-turbovote-signup [message]
+(defn-traced create-turbovote-signup [message]
   (turbovote-signup/handle notification/send-or-schedule message)
   {:status :ok})
 
 (defn delete-turbovote-signup [message]
   {:status :ok})
 
-(defn filter-subscription-mediums
+(defn-traced filter-subscription-mediums
```
tank157: I don't think we get anything by tracing this either.

src/election_notification_works/handlers/subscription_confirmation.clj
```diff
@@ -3,11 +3,12 @@
...
             [kehaar.wire-up :as wire-up]
             [clojure.core.async :as async]
-            [clojure.tools.logging :as log]))
+            [clojure.tools.logging :as log]
+            [com.climate.newrelic.trace :refer [defn-traced]]))
 
 (def read-user (wire-up/async->fn channels/user-read))
 
-(defn ->confirmation-notification-messages
+(defn-traced ->confirmation-notification-messages
```
tank157: Likewise, I wouldn't put tracing in here unless we were finding this call inside some code path that's taking a lot of time and we're adding lower level tracing to look for hotspots. I just can't see the couple of map merges and utility function calls being significant.

PR #74, opened 2016-09-22
Move scheduling emails to fire-and-forget

danielglauser:
The handlers in `election-notification-works` don't care if the email is sent in the next 40s or the next 10m, so stop timing out when things get backed up. This was identified as a bottleneck in the latest 400 person load test.


danielglauser:
This currently fails in Krakenstein, needs more testing.

danielglauser:
This be broken. Starting over.

PR #81, opened 2016-10-10
Tweaking the duration to prevent notifications

ddellacosta:
Temporarily (?)

...because we want to update some scheduled notifications today w/o sending duplicate notifications out.


ddellacosta:
Turned out we didn't use this approach. Closing.

PR #82, opened 2016-10-25
add a little more tracing

tank157:
Not strictly needed, but could be useful.


danielglauser:
You may want to break this up and add some additional tracing here: https://github.com/democracyworks/election-notification-works/blob/d2938708b5a40e18ddb3dc9de8ec162583753494/src/election_notification_works/polling_places.clj#L7
That way it's obvious which path the code is taking. If you don't want to and would just like this additional tracing out in prod, LGTM

tank157:
Well, I didn't touch that tracing yet, because we'll need to more or less re-write large chunks of that code to handle back-off retries, as well as to probably just hard code in the google election id as we've discussed before.

PR #84, opened 2016-10-30
dump out full stack trace when API call fails; don't try to call API for nil/blank zip

ddellacosta:
Updates for https://www.pivotaltracker.com/story/show/129806759


ddellacosta:
Again merging this pre-approval--will get some review after the fact.

PR #88, opened 2016-10-31
change sms UTM URL to have 'sms' string

ddellacosta:
does what it says on the tin

also see https://github.com/democracyworks/notification-templates/pull/32


danielglauser:
LGTM

PR #89, opened 2016-10-31
add retries for API calls

ddellacosta:
seems like we are killing the ETP voter guide API at the beginning when there are big bursts--adding some retries to hopefully slow it down a bit and make sure we don't lose anyone.


src/election_notification_works/handlers/voter_guide.clj
```diff
@@ -32,12 +32,24 @@
...
-                  (http/get voter-guide-api-url))]
-    (if (= 200 (:status resp))
-      (:body resp)
-      (log/error "Got a non-200 response: " (pr-str resp)))))
+  ([zip]
+   (call-external-voter-guide-api zip 10))
+  ([zip tries]
+   (try-try-again
+    {:decay :exponential :tries tries}
+    (fn [zip']
```
danielglauser: I don't think you need `zip'` here, `zip` will be in scope within the closure.

danielglauser: Not a big deal though, what you have here should work.

src/election_notification_works/handlers/voter_guide.clj
```diff
@@ -32,12 +32,24 @@
...
+                      (http/get voter-guide-api-url))]
+        (case (:status resp)
+          200
+          (:body resp)
+
+          500
+          (do (log/error "Got a 500 response, retrying: " (pr-str resp))
+              (throw (Exception. "Retry voter guide API call")))
+
+          (log/error "Got a non-200 or -500 response: " (pr-str resp)))))
```
danielglauser: There's no need to log here, you can use an [error hook](https://github.com/joegallo/robert-bruce/blob/master/test/robert/test/bruce.clj#L239).

danielglauser: https://github.com/joegallo/robert-bruce/blob/master/test/robert/test/bruce.clj#L239

danielglauser: No need, my bad. LGTM

ddellacosta: Per discussion in Slack think this may not apply here.

PR #94, opened 2016-12-13
Kehaar configured

tie-rack:
Another `kehaar.configured/init!` PR.


resources/config.edn
```diff
@@ -1,36 +1,135 @@
...
+    {:queue "electorate-works.election.election-users"
+     :channel election-notification-works.channels/electorate-election-users
+     :response true
+     :timeout 30000}
+    {:queue "election-works.election.read"
+     :channel election-notification-works.channels/election-read
+     :response true
+     :timeout 30000}
+    {:queue "election-notification-works.voter-guide.notify"
+     :channel election-notification-works.channels/voter-guide-notify-ext}]
```
danielglauser: Missing `:timeout 3000`. Needle in a haystack right there.

resources/config.edn
```diff
@@ -1,36 +1,135 @@
...
-                     "election-works.election.read" {:exclusive false :durable true :auto-delete false}}
-            :topics {"events" {:durable true :auto-delete false}}}
+{:rabbitmq
+ {:connection {:host #resource-config/env "RABBITMQ_PORT_5672_TCP_ADDR"
+               :port #resource-config/edn #resource-config/env "RABBITMQ_PORT_5672_TCP_PORT"}
+  :kehaar
+  {:incoming-services
+   [{:queue "election-notification-works.ok"
+     :f election-notification-works.handlers/ok
+     :response true}
```
danielglauser: Missing `:prefetch-limit 2`?

resources/config.edn
```diff
@@ -1,36 +1,135 @@
...
+     :prefetch-limit 10}]
+
+   :external-services
+   [{:queue "notification"
+     :channel election-notification-works.channels/notification-send}
+    {:queue "notification-works.notification.schedule"
+     :channel election-notification-works.channels/notification-schedule}
+    {:queue "notification-works.notification.unschedule"
+     :channel election-notification-works.channels/notification-unschedule}
+    {:queue "notification-works.notification.search"
```
danielglauser: Although there weren't `:timeout`s on all of these would it make sense to add them now? What happens if no timeout is specified?

resources/config.edn
tie-rack: That's a fire-and-forget service (`:response` isn't set), thus a timeout doesn't make sense.

resources/config.edn
tie-rack: The default timeout is 1000ms.... but you're right, these `:response true` services had timeouts before. Adding them back.

resources/config.edn
tie-rack: Good point. I'll limit it to 2 threads as well.

resources/config.edn
danielglauser: Nice, I missed that.

PR #98, opened 2017-02-06
Backend signup notifications

tank157:
Part of the NH work, moves the rendering of signup notifications to the backend, and the only piece of data that is scheduled is the user-id. Vastly simplifies the front-end, and makes it easier to reason about/update the contents of the notifications.


src/election_notification_works/handlers/signup.clj
```diff
@@ -0,0 +1,147 @@
...
+                                 scheduled/search-messages
+                                 scheduled/schedule-message)))
+
+(defn-traced event-electorate*
+  [search-fn schedule-fn {:keys [user]}]
+  (let [user-id (:id user)]
+    (log/debug "signup detected electorate event for " user-id)
+    (reschedule search-fn schedule-fn user-id)))
+
+(def event-eletorate
```
tank157: Fix typo

src/election_notification_works/vbm_form.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(def generate-vbm-form
+  (wire-up/async->fn channels/vbm-form-generate))
+
+(defn-traced generate
+  [electorate]
+  (when electorate
+    (if-let [response (async/<!! (generate-vbm-form {:electorate electorate
+                                                     :force false}))]
+      (do (log/debug "VBM RESPONSE:" (pr-str response))
```
tank157: Remove logging here

resources/config.edn
```diff
@@ -70,14 +78,39 @@
      :f election-notification-works.handlers/handle-user-delete
      :prefetch-limit 10}
     {:queue "election-notification-works.events.election-mail.subscription.create"
-     :routing-key "election-mail.subscription.create"
+     :routing-key "election-mail.subscription.*"
```
ddellacosta: I tried to find some comprehensive RMQ docs on wildcard routing-keys, but only found a stackoverflow answer suggesting that they don't exist.  What is this doing (what function does it have I mean) here and in other routing keys added to the config?

resources/config.edn
```diff
@@ -70,14 +78,39 @@
...
+     :routing-key "user.update"
+     :exchange "events"
+     :f election-notification-works.handlers.signup/event-user-update
+     :prefetch-limit 10}
+    {:queue "election-notification-works.events.signup.electorate"
+     :routing-key "electorate.*"
+     :exchange "events"
+     :f election-notification-works.handlers.signup/event-electorate
+     :prefetch-limit 10}
+    {:queue "election-notification-works.events.signup.voter-registraton-status"
```
ddellacosta: Typo?

ddellacosta: Although, does this even matter if this is an event queue listening for a specific routing-key(s), and just passes stuff along to the right handler?  That is, there's nothing really "on the other end" here, in the sense of anything referring directly to this queue, right?

resources/config.edn
cap10morgan: Wildcard routing keys _do_ exist. I'm having a hard time finding definitive docs on them, but they are at least discussed here: http://clojurerabbitmq.info/articles/exchanges.html#topic-exchanges

tank157: It was noticing that there were two event handlers doing the same thing here that encouraged me to try wildcards. There's a tiny sacrifice in that we don't log whether the event was a create or a delete specifically, but I think since the outcome is the same, it didn't really matter much.

Anyway, these just do the same thing as the other event handlers, pushing the signup notification a little later. But, thanks for pointing this out, the name of the queue should probably be adjusted.

src/election_notification_works/handlers/signup.clj
```diff
@@ -0,0 +1,147 @@
...
+                               (sub/read-subscription (dt/db))
+                               :mediums)]
+    (when subscriptions
+      (let [electorate (-> user-id electorate/electorates first)
+            reg-form-and-info (reg-form/generate user electorate)
+            vbm-form-and-info (vbm-form/generate electorate)
+            mailing-subscription (election-mail/user-subscribed? user)]
+        (when (and (contains? subscriptions :email)
+                   (contains? user :email))
+          (email/send-signup user reg-form-and-info vbm-form-and-info
```
ddellacosta: Style comment: once we get past two or three args I start to think a map is a better way to pass arguments in--it gets hard to remember order, for one.

EDIT: this is not a blocker, just a comment, take it or leave it.  If this was, like, eight args I might make more of an issue of it.

src/election_notification_works/handlers/turbovote_signup.clj
```diff
@@ -61,6 +61,7 @@
            (when (seq attachments) {:attachments attachments}))))
 
 (defn-traced handle*
+  "DEPRECATED: will be removed when new signup has been fully deployed."
```
ddellacosta: This and all other `DEPRECATED` notices: 👍

resources/config.edn
ddellacosta: Okay, thanks Wes and Troy for the response.

tank157: Ah, there we go. So this handler was previously two handlers, one for creation of election mail subs, and one for deletion, but both did the same thing. What they did was essentially update any scheduled notifications, but they didn't actually pass the new status of the election-mail subscriptions, the other message handlers looked them up again.

resources/config.edn
tank157: The typo won't really affect operation, but I'll fix it anyway.

ddellacosta: Right, that's what I thought, but thanks for the confirmation.

PR #100, opened 2017-02-22
Subscription export

tank157:
Needed for [this Pivotal Card](https://www.pivotaltracker.com/story/show/139393211)

Related [PR](https://github.com/democracyworks/turbovote-admin-works/pull/26)


tank157:
Assigning to @danielglauser

src/election_notification_works/handlers.clj
```diff
@@ -95,6 +95,29 @@
...
+    {:status :not-found
+     :user-id uuid}))
+
+(defn-traced page-subscriptions*
+  [message]
+  (log/debug "Handling page subscriptions for" (count message) "users")
+  (s/validate page-schema message)
+  {:status :ok
+   :subscriptions
+   (map (partial export-reader (db/db)) message)})
```
danielglauser: I realize the rest of this `ns` has mentions of `(db/db)` but I do want to point out that I don't think the `hander` `ns` should know about the database.

tank157: Good point, let me fix that.

PR #101, opened 2017-02-23
Fix overdelivery of vote-tomorrow notifications

tie-rack:
If a user signs up after the vote-tomorrow ideal send at, the notification may be sent multiple times as the user is updated. Using the same logic (but refactored) as the upcoming-election notification,
delay the sending of vote-tomorrow notifications to 10 minutes in the future if the ideal time isn't at least that far in the future.


test/election_notification_works/handlers/vote_tomorrow_test.clj
```diff
@@ -131,21 +131,21 @@
      (testing "email"
        (handle electorate #{:email})
        (let [message (:message (take-or-timeout channels/notification-schedule))
-             expected-send-after (DateTime. 2015 11 5 7 4 0 0 util/NY)]
+             expected-send-after #inst "2015-11-05T12:04"]
```
tank157: Could you maybe construct the #inst date in such a way that it highlights that the hour and minute match up with the configured values, with a timezone offset? I like how the previous implementation helps drive home that it's 7:04am Eastern.

tie-rack: Good call. Updated.

tank157:
👍

PR #104, opened 2017-04-03
deals with multiple subscriptions per-user by merging together subscriptions

ddellacosta:
At one point, we had a race condition and multiple subscriptions could be entered per user. We [resolved that issue](https://github.com/democracyworks/election-notification-works/pull/70) but apparently we still have cases where a user will have multiple subscriptions presumably from that old data still sitting around.

I've added [a card](https://www.pivotaltracker.com/story/show/142995105) to deal with cleaning the data up at some point but this PR should allow us to handle the data properly regardless, considering that we are currently seeing problematic differences in data from the export when compared to the dashboard.


src/election_notification_works/entities/subscription.clj
```diff
@@ -20,10 +19,13 @@
 (defn-traced read-subscription-entity
   ([user-id] (read-subscription-entity (db/db) user-id))
   ([db user-id]
-   (query/find-one '{:find [?subscription]
-                     :in [$ ?user-id]
-                     :where [[?subscription :user/id ?user-id]]}
-                   db user-id)))
+   (let [subs (d/q '{:find [[(pull ?e [*]) ...]]
```
tank157: I have a problem with doing the reduction here. It presents a fake entity (FAKE ENTITY! SAD!) to the caller, one that is not an actual representation of the database entity. And that could be used as the basis for other decisions (deletions, updates, etc). Could we change this to `read-subscription-entities` and then have the `read-subscription` do the reduction? That seems like it would be the right spot for translating between the internal versus external representation of the data.

tank157: While I do think this will work as is with `read-subscription` and `upsert-subscription*`, I don't think it'll work properly with the `delete-subscription` in handlers.clj.

ddellacosta:
Okay, after discussion we ended up agreeing that the right approach is actually to just fix the data. Going to close this one and move forward on that.

cc @tank157

PR #109, opened 2017-08-28
updates VBM form links to mostly point to our S3 bucket PDFs behind shortened links

ddellacosta:
https://www.pivotaltracker.com/story/show/141321671

Also adds dumb dev-src helper to (mostly) generate a ns from the CSV file; small refactoring of namespaces in `src/election_notification_works/handlers/upcoming_election.clj`.


dev-src/election_notification_works/vbm_links_csv_to_ns.clj
```diff
@@ -0,0 +1,26 @@
...
+(defn vbm-csv->ns
+  "Imports a CSV file matching https://docs.google.com/spreadsheets/d/1AlS6ZSzPUDWDD4_2v4RSfH6dZryUEdrupK3dvj3wT1M/edit#gid=0 and dumps a mostly massaged ns to `src/election_notification_works/absentee_ballots.clj`. Assumes we are working from root of repo."
+  [filepath]
+  (->> (csv/parse-csv (io/reader filepath))
+       (reduce #(assoc %1 (keyword (get %2 "State")) (get %2 "short url")) {})
+       (into (sorted-map))
+       into-ns-form
+       (spit "src/election_notification_works/absentee_ballots.clj")))
+
+(comment
```
marlabrizel: Did you mean to leave this in here? If so, is it worth changing this to be something more generic?

marlabrizel: The preview only shows the top of the `comment` form but what I'm referring to is actually the variable defined inside of the comment, not what's above it. Silly Github

ddellacosta: Yeah I left that there on purpose--mostly illustrates how it's used above, but didn't make sense to bake it in.

For `dev-src` code I have a lower standard--or maybe more accurately, I think env-specific details can be useful in `dev-src` code, as it is meant to be utility code for dealing with stuff that is relatively manual. So I tend to feel fine about including specific paths and whatnot.

See also https://github.com/democracyworks/turbovote-admin-works/blob/5b75c4e4c3aa855b03203692e6edf38e6bdea14d/dev-src/turbovote_admin_works/unsubscribe.clj#L122

marlabrizel: I see. How do you feel about adding a note like `;;Usage` or similar - like how it is in the example?

ddellacosta: Sure thing--let me know what you think of what I just updated.

PR #112, opened 2017-10-16
BugFix: add missing hostname to signup email ctx

marlabrizel:
**[Issue](https://www.pivotaltracker.com/story/show/152003083)**
The welcome email was not displaying a link for users to return to so that they could re-subscribe to TV in the event that they wished to update their registration.

**Fix**
Added a missing hostname key to the context available in the signup notifications.
Also took the opportunity to remove a redundant fn in order to standardize how we assoc context onto all of our emails.

Tested in Krakenstein


marlabrizel:
I don't have terribly strong opinions either way, but am inclined to leave it as is for the following reasons:
- Given the urgency of this bug fix (we're sending incomplete info to users at a high signup period), I want to push out sooner rather than later
- Given the painfully slow nature of testing anything related to notifications, I'm reluctant to delay pushing out a working solution

ddellacosta:
Okay--as I said I won't hold this PR up on account of this, but I _will_ point out that it's a very straightforward refactoring and its correctness should be confirmed by the current tests. In any case your call in the end.

cap10morgan:
OK, now that this fix is headed to production, I'd be happy to address your refactoring request, Dave. Part of the issue yesterday is that Marla and I preferred certain aspects of the code that was already in place but didn't feel it was worth the time to get into it as blocker for that bug fix, so we took you up on your "I won't let this hold up the PR" offer (and thank you for that).

One thing I _do_ really like in your suggested refactoring is the cleanup of the `full-state-name` calculation. The nested `get`s are kind of hard to follow as-is. But it struck me as a good candidate for another function extraction like the other elements of this map use.

Do you have any lingering concerns over this refactor?

```clojure
(defn registered-state-name
  "Returns the full unabbreviated name of the state in which `user` is or will
  be registered to vote."
  [user]
  (let [state-abbreviation (get-in user [:addresses :registered :state])]
    (get state-names state-abbreviation)))

(defn user->recipient
  "Format a `user`'s contact information to turn them into a notification
  recipient data structure."
  [user]
  (-> user
      (assoc :first-name (preferred-first-name user))
      (assoc :full-name (full-name user))
      (assoc :state (registered-state-name user))))
```

The tests still all pass after this change.

ddellacosta:
@cap10morgan Nope, looks good to me. cc @marlabrizel

PR #116, opened 2017-10-25
Add authority polling place url

ddellacosta:
UPDATE: this PR now _also_ adds a last-minute authority lookup for ensuring we are getting the most current data for polling-place-urls and voting-id in-person voting method instructions from the authority. Relevant card: https://www.pivotaltracker.com/story/show/152485111

Tracker card: https://www.pivotaltracker.com/story/show/141444789

Relevant PRs:
https://github.com/democracyworks/election-authority-works/pull/49
https://github.com/democracyworks/election-works/pull/62
https://github.com/democracyworks/notification-templates/pull/54


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -52,114 +50,135 @@
...
   (log/debug "vote-tomorrow handling an electorate:" (pr-str electorate))
-  (let [election (:election electorate)]
-    (when-not (too-late-for-notification? election)
-      (let [notifications (->vote-tomorrow-notification-messages mediums
-                                                                 electorate)]
-        (doseq [n notifications]
-          (notification/send-or-schedule n))))))
+  (when-not (too-late-for-notification? election)
+    (let [notifications (->vote-tomorrow-notification-messages mediums electorate)]
+      (doseq [n notifications]
```
gws: This could be `(run! notification/send-or-schedule notifications)`

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -1,22 +1,21 @@
...
             [election-notification-works.notification :as notification]
-            [turbovote.resource-config :refer [config]]
-            [com.climate.newrelic.trace :refer [defn-traced]])
-  (:import [org.joda.time DateTimeZone DateMidnight DateTime]
-           [java.net URLEncoder]))
+            [election-notification-works.polling-places :as polling-place]
+            [election-notification-works.voting-methods :as vm]
+            [turbovote.resource-config :refer [config]])
+  (:import [java.net URLEncoder]
+           [org.joda.time DateTimeZone DateMidnight DateTime]))
```
gws: When we reorder `require`s and `import`s, do we also want to re-indent them like we had started doing? E.g.

```clojure
(:require
 [... :as ...])
(:import
 (java.foo Whatsit)
```

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -52,114 +50,135 @@
...
 (defn map-url
-  "Takes a `destination` address string (e.g.
-  \"123 Main St Springfield MA 55555\") and returns a Google Maps URL to show
-  directions from a default location (usually the user’s current location) to
+  "Takes a `destination` address string (e.g. '123 Main St Springfield
+  MA 55555') and returns a Google Maps URL to show directions from a
+  default location (usually the user’s current location) to
   `destination`.
-
+
```
gws: extra space

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: @gws I don't think we ever came to a consensus so I've simply been sorting them to ensure I eliminate dupes and reduce cognitive load, and otherwise leave them as is. Honestly not even sure if everyone ended up agreeing with the idea to alphabetize them.

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: Cool, didn't know about that function. Will update.

src/election_notification_works/handlers/vote_tomorrow.clj
gws: Leaving them indented where they are helps with consistency across the code, anyway.

:+1:

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: Oh you know, I added this space it in as I thought it made the docstring more readable. Disagree?

gws: Oh, I like the newline that was there, I'm only talking about the leading space that just looked like an editor quirk when reformatting a block of text.

Original:

https://github.com/democracyworks/election-notification-works/blob/master/src/election_notification_works/handlers/vote_tomorrow.clj#L75-L77

ddellacosta: Oh okay, I misunderstood/misremembered I guess. Will clean that up.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -80,101 +71,134 @@
...
+    (let [address-string (map->string
+                          (:address polling-place)
+                          [:line1 :line2 :line3 :city :state :zip])]
+      (->> (map-url address-string)
+           (assoc polling-place :map-url)
+           (assoc context :polling-place)))
+    context))
+
+(defn assoc-polling-place-urls-if-exists
+  "Take a context and an election, and if a polling-place-url is present in the election, associates the polling-place-url with the context and returns the context,otherwise returns the context with no modification."
```
zonotope: line wrap please

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -185,3 +209,47 @@
...
+     (let [{:keys [status authorities] :as wtf} (search-fn state-ocd-id)]
+       (if (= :ok status)
+         (do (swap! cache assoc state-ocd-id (first authorities))
+             (first authorities))
+         (throw (Exception. "No authority returned for " state-ocd-id)))))))
+
+(defn state-ocd-id
+  [ocd-ids]
+  (->> (filter #(re-find #"state" %) ocd-ids)
+       (map #(str/replace % #"^(ocd-division/country:us/state:\w{2,2}).*$" "$1"))
```
tank157: Is there a story behind this construct? It seems like you could filter/re-find with the longer regex, but just with the `.*` removed from the end?

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -185,3 +209,47 @@
...
+
+(def authority-cache (atom {}))
+
+(defn get-state-authority
+  ([search-fn state-ocd-id]
+   (get-state-authority authority-cache search-fn state-ocd-id))
+  ([cache search-fn state-ocd-id]
+   (if-let [authority (get @cache state-ocd-id)]
+     authority
+     (let [{:keys [status authorities] :as wtf} (search-fn state-ocd-id)]
```
zonotope: wtf? 😉

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -80,101 +71,134 @@
...
-                                                           user-reg-address))
-                               notification)]
+  [authority-search-fn {{:keys [recipient election]} :context, :as notification}]
+  (let [registered-addr (get-in recipient [:addresses :registered])
+        update-fn (fn [context]
+                    (-> context
+                        (assoc-polling-place-urls-if-exists election)
+                        (assoc-polling-place-if-exists
+                         registered-addr (:google-election-id election))
+                        (temporary-update-authority-data authority-search-fn)))
```
tank157: Part 1 of 2:

This is a 2 part comment: the temporary-update-authority-data update is in the update workflow here.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -80,101 +71,134 @@
...
+  constructed from the `user` and `election` present in the
+  `electorate`."
+  [mediums {:keys [user election] :as electorate}]
+  (let [registered-address (get-in user [:addresses :registered])
+        context (-> {:recipient (util/user->recipient user)}
+                    (assoc :election (electorate->ctx electorate))
+                    (assoc :signers (util/signature))
+                    (assoc-polling-place-if-exists
+                     registered-address (:google-election-id election))
+                    (assoc-polling-place-urls-if-exists election))]
```
tank157: Part 2 of 2:

But it's not in this place, which does all the other things too. Now, maybe that's ok, we won't need to update the authority info if this is essentially going out immediately (ie, it's skipping the update because the user is literally created on the day before the election). Is that the reason it's left out of here?

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -185,3 +209,47 @@
...
+
+(defn get-state-authority
+  ([search-fn state-ocd-id]
+   (get-state-authority authority-cache search-fn state-ocd-id))
+  ([cache search-fn state-ocd-id]
+   (if-let [authority (get @cache state-ocd-id)]
+     authority
+     (let [{:keys [status authorities] :as wtf} (search-fn state-ocd-id)]
+       (if (= :ok status)
+         (do (swap! cache assoc state-ocd-id (first authorities))
```
zonotope: you repeat `(first authorities)` here. i'd wrap this in a `let` instead of a `do` to avoid that:

```clojure
(if (= :ok status)
  (let [authority (first authorities)]
    (swap! cache assoc state-ocd-id authority)
    authority)))
```

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: Yeah, good question. This will _only_ get picked up when we hit the update endpoint, since that's where these vote-tomorrow messages go when they are actually sent out. Because I've already updated the (state) authorities today, any folks who sign up for notifications between then and now will actually already have the updated info in here--but regardless they should go through the update handler as well before finally being sent out. Does that answer your question adequately?

src/election_notification_works/handlers/vote_tomorrow.clj
zonotope: i think finding with the longer regex directly will work by itself, and be a lot easier to understand (at least for me).

src/election_notification_works/handlers/vote_tomorrow.clj
tank157: Yeah, sure does. Just wanted to make sure it was all intentional/thought out!

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: So the `filter` operation first isolates all ocd-ids that have a state-level value embedded, then the `map` is just there to let us actually get the state ocd-id from the first one of ocd-ids the `filter` produces. I'm not sure how I'd compress these into one operation in the way you're suggesting, although maybe I'm not understanding it--`replace` simply gives me the string back when it doesn't match, so it can't be used to filter, and conversely `re-find` won't let me do a replace to isolate the state-level ocd-id, as far as I can see.

ddellacosta: Maybe this would be more clear if it was

```clojure
(-> (filter #(re-find #"state" %) ocd-ids)
    first
    (str/replace #"^(ocd-division/country:us/state:\w{2,2}).*$" "$1"))
```

? It's semantically equivalent because of the laziness of the map.

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: Oh whoops, meant to do that...

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: I mean, personally I think it's an extra line with little benefit other than to adhere to the dogma of DRYness, but I don't care that much.

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: Whoops!!! I guess you can tell that something wasn't working at some point...

src/election_notification_works/handlers/vote_tomorrow.clj
ddellacosta: I could also use the same regex in both places, I just figured `state` was enough for the initial filter to find all the ocd-ids that I could pull the state ocd-id out of.

tank157: If I'm understanding the intent correctly, you want the OCD-ID that is just at the state level. All the OCD-IDs have "state" in them, with the exception of the one that is just the country level, so the first filter isn't really doing much. You end up with the one state level OCD-ID, county level OCD-ID, city level OCD-IDs, etc. Then the replace is using regex match grouping to pull just the part of each of those that is the state level component. So from `ocd-division/country:us/state:co/county:denver` it's doing to match `ocd-division/country:us/state:co` and replace the whole string with just that part?

Instead, you could do `(first (filter #(re-find #"^ocd-division/country:us/state\w{2,2}$" %) ocd-ids))` I think. It would find the one that's at the state level only and return that, no regex group matching/string replacement needed.

ddellacosta: Oh gotcha--I wasn't sure I could always assume the state-level ocd-id would be present. That makes sense Troy, thanks for explaining it in detail, I get it now. It does indeed work exactly as you laid it out (with the addition of a colon):

```clojure
election-notification-works.dev-utils> wi-ocd-ids
#{"ocd-division/country:us" "ocd-division/country:us/state:wi/cd:4" "ocd-division/country:us/state:wi/county:milwaukee" "ocd-division/country:us/state:wi" "ocd-division/country:us/state:wi/place:milwauke\
e"}
election-notification-works.dev-utils> (first (filter #(re-find #"^ocd-division/country:us/state:\w{2,2}$" %) wi-ocd-ids))
"ocd-division/country:us/state:wi"
election-notification-works.dev-utils>
```

tank157: Something would be pretty wrong from our OCD-ID source if they left the state out. I'm pretty sure we rely on it being there in other places. Though, while I think about it, what does DC look like? It might be a special case.

ddellacosta:
Thanks a lot @zonotope and @tank157 for the additional comments and critiques. I'm going to do a bit more testing of this in krakenstein, then push it to production.

src/election_notification_works/handlers/vote_tomorrow.clj
tank157: Google has this for DC:
```{
 "kind": "civicinfo#divisionSearchResponse",
 "results": [
  {
   "ocdId": "ocd-division/country:us/district:dc",
   "name": "District of Columbia",
   "aliases": [
    "ocd-division/country:us/state:dc"
   ]
  }
```

ddellacosta: Good catch...gonna have to handle that specially I guess. Thanks very much Troy!

tank157: Yeah, now that I think about it, this is why I wanted to change things to send _all_ of a user's ocd-ids to `election-authority-works`. Because now we have another place we're going to have custom code to find a state ocd-id and can easily forget about the special cases like this. No critique on this work in particular, just we need to consolidate this algorithm into one place.

ddellacosta: Absolutely agreed.

tank157: Oh, I just missed that alias for `state:dc`, since we merge those in now, I think the state search would work. Test it out in Krakenstein to make sure

ddellacosta: @tank157 @zonotope Okay just pushed a new version--let me know if you think this has any remaining issues.

ddellacosta: It doesn't for the authority search, it would seem. :-/ At least it didn't locally. In any case think it's just as well to leave this special check in to be safe, especially since the intention is to remove this code at some point soon after the election.

tank157: Yeah, I went and looked at the spreadsheet next, we have the "state" ocd-id for DC listed as `ocd-division/country:us/district:dc`, so the alias wouldn't have helped.

PR #117, opened 2017-11-22
add compact response for `page-subscriptions*`

ddellacosta:
Adds a more efficient way to retrieve a batch of subscriptions with a more compact response by augmenting the current `page-subscriptions*` handler with an extra arg (and hackily coercing the vector format to be a map, which I hope to remove soon-ish). Calls to the `page-subscriptions*` handler using a vector of user-ids should continue to work as expected (used in `turbovote.admin.works.handlers.export`).

See
https://www.pivotaltracker.com/story/show/151059862

And related:
https://github.com/democracyworks/turbovote-admin-http-api/pull/43
https://github.com/democracyworks/turbovote-admin-web/pull/55


src/election_notification_works/handlers.clj
```diff
@@ -109,11 +115,19 @@
...
 (defn-traced page-subscriptions*
   [message]
-  (log/debug "Handling page subscriptions for" (count message) "users")
   (s/validate page-schema message)
-  {:status :ok
-   :subscriptions
-   (map export-reader message)})
+  ;; We should modify the export (turbovote-admin-works) to send a map
+  ;; as a request vs. vector, then this can be removed. But leaving
+  ;; that for the time being as that's out of scope for the work I'm
```
zonotope: this is no biggie, but i would re-word this comment because by the time this hits master, you won't be doing the work "now" anymore. 

instead, i'd say something like
```clojure
;; TODO: remove vector based dispatch path once `turbovote-admin-works` is
;;       modified to send a map request payload
```

PR #118, opened 2017-12-07
Fix errors when running exports in the admin console

ddellacosta:
This was caused by the schema validation only checking for vectors when in fact the export's message format was a list of user-ids. So this PR makes the schema check more general to ensure we handle vectors and lists appropriately, and changes the tests to be lists to reflect what we actually expect the export to send us. This should still handle vectors fine if need be.

Tested this in krakenstein as well, for both exports and (the other use-case for this endpoint) loading subscriptions for user search results in the admin console.


ddellacosta:
@zonotope I also adjusted the comment (with a few tweaks of my own) to reflect your suggestion in the last PR which I inadvertently blew off--sorry!

PR #120, opened 2018-01-16
added formatter for Spanish date and check for language before returning formatted date

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/153297378) I added another formatter in [election-notification-works.handlers.util](https://github.com/democracyworks/election-notification-works/compare/election-date-spanish?expand=1#diff-f6fe2adf838630d606e7f972bc5ccaf1) to account for formatting in Spanish as well as in English and added another if in the `->election` function to check the language preference and select the formatter accordingly.  There's also a test in [election-notification-works.handlers.upcoming-election-test](https://github.com/democracyworks/election-notification-works/compare/election-date-spanish?expand=1#diff-1bb9671cc4596c68849b0873c739eca6) to confirm that when the language preference is `:es` that the date will be added to the context as expected.


test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -330,6 +330,41 @@
...
+                              :form-uri "rendered-form-uri"
+                              :wants-vbm? true
+                              :has-form? true
+                              :authority-lookup "authority-lookup-url"
+                              :authority (:election-authority electorate)})
+
+      (handle-update notification)
+
+      (let [message (:message (take-or-timeout channels/notification-send))]
+        (is (= (-> message :context :election :date) "sábado, el 1 de  febrero")))))
```
cap10morgan: Can we get rid of the extra space between `de` and `febrero`?

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -24,11 +24,14 @@
 
 (defn ->election
   "Format information about an `election` for the requested medium."
-  [election medium]
+  [election medium language-map]
```
cap10morgan: Are other functions calling this in a way that makes sense for this to be a map? In this code (which I realize was already this way before you touched it), it looks like this is put into a single-key map only to pull it right back out again in here.

If the code below is the only thing calling this, let's just make `language-map` a simple keyword arg and just pass the user's language-preference.

src/election_notification_works/handlers/util.clj
```diff
@@ -6,7 +6,8 @@
             [clojure.tools.logging :as log]
             [clojure.core.async :as async]
             [election-notification-works.channels :as channels]
-            [kehaar.wire-up :as wire-up]))
+            [kehaar.wire-up :as wire-up])
+  (:import [java.util Locale]))
```
cap10morgan: Imports should use lists instead of vectors (though this is a matter of idiomatic debate I suppose). So like this:
`(:import (java.util Locale))`

src/election_notification_works/handlers/util.clj
```diff
@@ -195,6 +196,10 @@
 
 (def email-date-formatter (fmt/formatter "EEEE, MMMM d"))
 
+(def en-email-date-formatter (fmt/formatter "EEEE, MMMM d"))
+
+(def es-email-date-formatter (fmt/with-locale (fmt/formatter "EEEE, 'el' d 'de'  MMMM") (Locale. "es")))
```
cap10morgan: Hard to see here, but looks like the extra space I refer to below is between `'de'` and `MMMM`.

src/election_notification_works/handlers/upcoming_election.clj
katrinamariehh: `language-map` is used at the bottom of the `->upcoming-election-function` (it gets merged with the message) and I was trying not to mess with too much stuff--I could pull the language out of the map on the call to `->election` and then have that function just expect the language keyword, but there's also a possibility that there's _no_ language preference (i.e. I think then the value of `language-map` is either `nil` or `{:language nil}`, but then I think it will just pass `nil` and be fine then?  I think I'll go with that--change the function to expect just the language keyword and pull it out when I make the call to `->election` inside of `->upcoming-election-function`.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -145,7 +144,7 @@
         recipient (util/user->recipient user)
         language-map (when-let [language (:language-preference user)]
                        {:language language})
-        election (->election (:election electorate) medium language-map)
+        election (->election (:election electorate) medium (:language language-map))
```
cap10morgan: This is looking better.

Instead of pulling this out of the map here, we could just not put it in a map in the first place on line 145/146. That will requiring changing this line and line 163/164 too. I'm not seeing any need to have it in a map.

So the 3 lines in question could become something like:
```clojure
language (:language-preference user)
```

```clojure
election (->election (:election electorate) medium language)
```

```clojure
(if language (assoc message :language language) message)
```

PR #122, opened 2018-01-24
Address confirmation message

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/153296925) I've edited the flow for `upcoming-election` and `vote-tomorrow` SMS notifications to include an additional message with details about the address the message relates to.  There are also tests in place to confirm the additional message shows up.


test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -524,22 +524,30 @@
...
              second-message (:message (take-or-timeout channels/notification-schedule))
+             third-message (:message (take-or-timeout channels/notification-schedule))
              email-message (first (filter #(= :email (:medium %))
-                                          [first-message second-message]))
+                                          [first-message second-message third-message]))
              sms-message (first (filter #(= :sms (:medium %))
-                                        [first-message second-message]))]
+                                        [first-message second-message third-message]))
+             sms-message2 (second (filter #(= :sms (:medium %))
+                                          [second-message third-message]))]
```
zonotope: i would also include the `first-message` in the list here you're filtering over just in case order is not preserved.

katrinamariehh:
Thanks for all the suggestions @zonotope!  I have read through the style guide but it's been a while, thanks for taking the time to point out specific places I can improve!  I'll get these changes made ASAP.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -293,6 +317,15 @@
             :identifiers {:recipient-id (get-in electorate [:user :id])
                           :election-id (get-in electorate [:election :id])}})))
 
+(defn ->create-followup-address-message
```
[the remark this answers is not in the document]
katrinamariehh: I *definitely* didn't know this is what the arrows were supposed to mean!  I think I've mostly used them in mimicking other places where they've been used, but this is super helpful to know.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -114,6 +114,29 @@
                      (not (get-in context [:recipient :should-request]))))]
     (assoc-in context [:recipient :paper-vbm-form] status)))
 
+(defn electorate->address-followup-message
+  [{:keys [user election-authority] :as electorate}]
+  [electorate]
+  (let [recipient (util/user->recipient user)
+        election (->election (:election electorate) :sms)
+        election-authority (->election-authority election-authority)
```
katrinamariehh: @zonotope one of the changes means we now have this line, which feels weird but is maybe OK?  What do you think?

zonotope: Is repeating the `election-authority` binding the thing that feels weird, or is it something else?

If it's repeating the binding, then yes, that's something I also try to avoid. The code still works as long as you don't need the original `election-authority` value, but it could be confusing.

If it were me, I would probably call the new value (the result of calling `(->election-authority election-authority)`) something like `authority` to avoid clobbering the original value.

src/election_notification_works/handlers/upcoming_election.clj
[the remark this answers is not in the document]
zonotope: yeah, that kind of thing is hard to tell from just reading code.

src/election_notification_works/handlers/upcoming_election.clj
katrinamariehh: I think it's the repeat of `election-authority`, your suggestion to re-name will make it seem less weird, I think!

PR #123, opened 2018-01-29
Provide registration authority address with reg form SMS

katrinamariehh:
For [this card](https://www.pivotaltracker.com/n/projects/930636/stories/154548051) I added another message after the `:voter-reg-pdf` message with details about the election authority's address the voter registration from should be mailed to.

Works in conjunction with [this PR on notification-templates](https://github.com/democracyworks/notification-templates/pull/60).


cap10morgan:
Merging in master should fix that build failure.

PR #124, opened 2018-01-31
Display election deadlines in Spanish for spanish-language notifications

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/154650115) I've updated the formatting for `deadlines.registration.date` and `deadlines.vbm.date` to check the language preference of the user and format the date in Spanish if their language preference is Spanish.  I've also updated the tests to make sure dates are coming through in both English and Spanish as expected for these dates as well as the `election.date`.


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -37,7 +37,7 @@
 (defn ->deadlines
   "Calculate the deadlines for voter registration and voting by mail
   in a given `electorate`, formatted for the requested medium."
```
cap10morgan: We should add `and language` to the end of that docstring.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -49,12 +49,19 @@
...
+                         (fmt/unparse-local-date util/es-email-date-formatter))}
+         :vbm
+         {:date (some->> vbm-deadline
+                         (fmt/unparse-local-date util/es-email-date-formatter))}}
+        {:registration
+         {:date (some->> voter-reg-deadline
+                         (fmt/unparse-local-date util/en-email-date-formatter))}
+         :vbm
+         {:date (some->> vbm-deadline
+                         (fmt/unparse-local-date util/en-email-date-formatter))}}))))
```
cap10morgan: It would be better to DRY this up a little more. Could we do something like set the formatter based on the language and then only have one instance of the `{:registration ..., :vbm ...}` map?

katrinamariehh: Yeah, it's definitely a little verbose here, I'll see what I can do to make it nicer!

test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -330,41 +330,85 @@
...
-                             :has-form? true
-                             :authority-lookup "authority-lookup-url"
-                             :authority (:election-authority electorate)})
-
-     (handle-update notification)
-
-     (let [message (:message (take-or-timeout channels/notification-send))]
-       (is (= (-> message :context :election :date) "sábado, el 1 de febrero")))))
-  
+   (testing "Sending an email notification with Spanish language-preference gets expected date format"
```
cap10morgan: Looks like this whole `testing` form got indented 3 spaces instead of 2 (like the rest of the file), which makes the diff hard to see in here. Can you scoot it back 1 space?

katrinamariehh:
I think I got the indentation fixed -- I opened it in atom to fix the merge conflicts and I think my auto-formatting there might have been the problem, lmk if this works.

PR #128, opened 2018-03-12
Add registration lookup urls to notification context

theJohnnyBrown:
Allows a link to the registration status in the upcoming election notification.


src/election_notification_works/entities/authority.clj
```diff
@@ -0,0 +1,34 @@
+(ns election-notification-works.entities.authority
```
zonotope: i think i'm on board with refactoring this code to it's own namespaces, but i don't think it belongs under `entities`. we typically use entities to refer to code meant to manipulate data we write to and read from datomic. 

i'd put this namespaces either top-level, directly under `election-notification-works`, or in another module like `election-notification-works.client` or something like that.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -169,10 +170,18 @@
   [electorate medium]
   (let [user (:user electorate)
         recipient (util/user->recipient user)
+        ocd-id (->> (get-in recipient [:addresses :registered :ocd-divisions])
+                    util/state-ocd-id)
+        state-authority (get-state-authority ocd-id)
+        state-ocd-reg-lookups (-> state-authority
+                                  (get-in [:ocd-division 2])
```
zonotope: this `get-in` call here is a little opaque. perhaps it would be clearer if we added a function to the `authority` namespace with a descriptive name that parses out whatever data is under the `[:ocd-division 2]` path of the state authority, and then we could call that function here and continue with the `select-keys ...` call below. i think that will make this code easier to follow along with from this namespace because that function name would be more descriptive than `(get-in [:ocd-division 2])`.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -181,6 +190,10 @@
                  :medium medium
                  :context {:recipient recipient
                            :election election
+                           :voter-registration-status-url
```
zonotope: is there a reason why we don't add both the `voter-registration-status-url` and the `voter-registration-status-url-shortened` to the context and let the template sort out which url should be displayed?

theJohnnyBrown: No reason, I'll add both

src/election_notification_works/entities/authority.clj
theJohnnyBrown: Good to know. IMO it should be in its own namespace now that `upcoming-election` and `vote-tomorrow` are both using it, because the alternative of having one handler depend on another feels risky.

src/election_notification_works/handlers/upcoming_election.clj
theJohnnyBrown: I guess I wrote it so it's clear to me, but a function such as `->ocd-division-map` that you have to jump around to read is more complexity than looking at the code in the arrow macro and seeing that it pulls the attributes map out of an ocd-division vector (which are used all over the code wherever it handles ocd-divisions). I disagree but I'll keep an open mind if you feel strongly

theJohnnyBrown: On second thought I think the clearest thing would be something like 

`(let [...
          [reg-lookup-url reg-lookup-url-shortened] (registration-status-urls state-authority)
          ...
          ]
     message {... }
)

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -243,50 +239,16 @@
 ;; FROM HERE BELOW IS PROVISIONAL CODE TO WORKAROUND HAVING TO UPDATE
```
ddellacosta: If this is going to become more or less permanent code I guess this comment should go away?

src/election_notification_works/authorities.clj
```diff
@@ -0,0 +1,42 @@
...
+            [kehaar.wire-up :as wire-up]
+            [clojure.tools.logging :as log]
+            [clojure.core.async :as async]))
+
+(defn authority-search-fn
+  [ocd-id]
+  (let [search-fn (wire-up/async->fn channels/authority-search)]
+    (async/<!! (search-fn {:ocd-divisions [ocd-id] :language :all}))))
+
+(def authority-cache (atom {}))
```
ddellacosta: If this is going to be permanent code, it probably makes sense to replace this with a real cache that doesn't require us restarting the service to clear out--what about something like an [LRU cache](https://github.com/clojure/core.cache/wiki/LRU) from `core.cache`?

theJohnnyBrown:
@ddellacosta my understanding is that we need an authority lookup because these URLs are stored on the state-level authority entity and often the `election-authority` and `registration-authority` already present in the electorate will be the county-level authority, which lacks that information.

ddellacosta:
@theJohnnyBrown I think that information _should_ get pulled in when the election is imported initially and these notifications are generated, but I'm forgetting the details. I'm asking because I do remember that the main reason I added the code before the last election was that it was simply easier than figuring out how to reschedule a bunch of notifications (the links I added weren't present at the time the elections were imported, so that data was not pulled and stored in `scheduled-message-works`). It was a bit ambiguous whether or not we would want to do this in general, however, since it's somewhat costly without the very stupid atom cache I added. @cap10morgan may have more to say here too.

More context [here](https://github.com/democracyworks/election-notification-works/pull/116/files) too.

src/election_notification_works/handlers/upcoming_election.clj
zonotope: yeah, the part that i thought was opaque is the `[:ocd-division 2]` path. what's supposed to be under 2? what about 1? the index of the vector doesn't convey any information about what the code expects to find there. contrast that with a map key that is usually a semantic name for the data the value contains

i was worried that at some point these vector indices could shift and this code breaks because of that. in that case, it would be more difficult to figure out what this code expects to find under index 2, and some sort of semantic name (like a map key or well-named parsing helper function) would be more explicit about what the code expects and will point to what needs to be fixed to accommodate the shifting vector.

src/election_notification_works/authorities.clj
gws: Sorry, I posted my comment before seeing this! I would probably advocate instead for a [TTL cache](https://github.com/democracyworks/voter-registration-works/blob/master/src/voter_registration_works/election_authorities.clj#L11-L32) but I *definitely* agree that it should not be "forever" (until the next restart).

ddellacosta: Yeah, TTL makes sense to me too. In any case, agreed that whatever it is that atom should go away!

theJohnnyBrown: Plus one to the TTL cache, I think we can fit all 51 state/district authorities in memories so we don't really need to evict for space

ddellacosta: BTW @gws now that I think about it I think I'd still push for the LRU based on the way this data will be access--most likely we'll see a bunch of messages going out and all needing the same authority's data centering around a specific point in time. This being the case it made the most sense to me that we'd simply drop anything old on the floor, as that reflects how the data is going to be used. BUT I'm interested in hearing why you think TTL is the better fit here--I may be missing some aspect of how that supports the data access pattern better here.

ddellacosta: @theJohnnyBrown The thing is, we don't care much about holding on to this data I don't think--the main thing we want to prevent is 10,000 hits (or whatever) to `election-authority-works` all at once, but a very small LRU cache size combined with the fact that we are only ever going to need a few authorities' data at the same time means we can always have the latest without concerning ourselves with anything other than eliminating a massive number of requests for authority data happening during sending.

theJohnnyBrown: Makes sense - How about an LRU with, say, 16 entries?

ddellacosta: @theJohnnyBrown I think that's more than enough, we can also tune it if we see performance drop-offs for some reason.

I'm still curious if @gws has some counterpoints too...

ddellacosta: Actually I have my own counterpoint: I'm not _positive_ that the only access pattern we have to contend with is big bursts of activity all hitting a single authority for a period, followed by another burst hitting a different authority--we may get pathological behavior with the LRU cache if we have a lot of user signups trickling in, with certain states represented more than others. That being the case I may back away from my suggestion and agree with Gordon--it's possible a TTL will more reliably let us cache the authority data and perform effectively without this data getting too stale.

Yeah, caching is hard.

theJohnnyBrown: We are talking about less than a megabyte here

theJohnnyBrown: I'm going to go ahead with the LRU and add a log statement so we can investigate any spike in cache misses that may occur

ddellacosta: To be clear, I'm less worried about the size of the thing we're caching than how it updates and when.

gws: Yes Dave I am thinking along the lines of your most recent comment. With LRU, you're leaving it up to access patterns to determine when data is evicted. Suppose you tell the LRU cache to save the 10 most recently accessed authorities - in situations where only 10 different states are accessed, they will be stale for an arbitrary amount of time. It would require 10 completely different states to be accessed in order to evict the 10 that had been cached.

Setting the LRU cache policy lower, say, to 1 would just result in extra thrashing as the single state is being evicted and replaced with a new state if even two authorities are being accessed, and so on.

One main benefit of an LRU cache, an upper bound on how much space it uses, is not desirable in our case.

A TTL cache would give us a "maximum staleness" window that we could rely on *regardless of access pattern*, and the type of data we use with the cache already places an upper bound on the amount of space this cache will occupy.

ddellacosta: @theJohnnyBrown so, not to beat this dead horse endlessly, but FWIW seems like @gws and myself now both believe the TTL cache is the way to go.

theJohnnyBrown: On the bright side these tests are doing a good job of keeping the interface consistent :)

gws: I'm sorry if it feels like I dragged everyone out to the bikeshed - ultimately either of these options are technically an improvement over what was there :+1:

theJohnnyBrown:
@ddellacosta @zonotope if I've addressed your comments could I get a ✅?

theJohnnyBrown:
Never mind there's some more test failures

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -243,50 +238,16 @@
...
-   (first (filter #(re-find #"^ocd-division/country:us/district:dc$" %) ocd-ids))
-   (first (filter #(re-find #"^ocd-division/country:us/state:\w{2,2}$" %) ocd-ids))))
-
-(defn maybe-assoc-in
-  [m keypath value]
-  (if (seq value)
-    (assoc-in m keypath value)
-    m))
-
 (defn temporary-update-authority-data
```
ddellacosta: Also...maybe this name should no longer be `temporary-`...?

theJohnnyBrown:
Isolating tests that deal with mock atoms and channels got me like 

mind blown

ddellacosta:
@theJohnnyBrown I hear that.

Will take another look now.

src/election_notification_works/authorities.clj
```diff
@@ -0,0 +1,65 @@
...
+(defn authority-search-fn
+  [ocd-id]
+  (let [search-fn (wire-up/async->fn channels/authority-search)]
+    (async/<!! (search-fn {:ocd-divisions [ocd-id] :language :all}))))
+
+(defn authority-lookup [state-ocd-id]
+  (log/warn "Cache miss for state election authority: " state-ocd-id)
+  (let [{:keys [status authorities]} (authority-search-fn state-ocd-id)]
+    (if (= :ok status)
+      (let [authority (first authorities)]
```
ddellacosta: This could just be `(first authorities)` I think?

src/election_notification_works/authorities.clj
```diff
@@ -0,0 +1,65 @@
...
+            [clojure.core.async :as async]
+            [clojure.core.cache :as cache]))
+
+(def authority-cache (atom (cache/lru-cache-factory {} :threshold 16)))
+(defn clear-state-authority-cache! []
+  (reset! authority-cache (cache/lru-cache-factory {} :threshold 16)))
+
+
+(def ^{:private true} default-wrapper-fn #(%1 %2))
+(defn through-cache
```
ddellacosta: I  f^#%ing hate that you had to do this but not sure what the alternative is. `nil` is evil. Oh well.

src/election_notification_works/authorities.clj
theJohnnyBrown: Thanks! much neater

PR #129, opened 2018-03-13
Ballot preview

zonotope:
### Description 

This adds ballot preview data to the context to be rendered by notification templates. In order to do this, I've refactored the previous civic-info requests for polling locations to make them more general and return the full google civic info response, and later on pull both ballot info and polling places from that response. This means we can get all the data we need with one civic info api request instead of multiple. 

### Feature Implemented

https://www.pivotaltracker.com/story/show/153568551


zonotope:
@cap10morgan they are not required, but only if there is an unambiguous upcoming election for the address. i believe that's the only use case we care about here.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -176,13 +168,16 @@
   `electorate`."
   [mediums {:keys [user election] :as electorate}]
   (let [registered-address (get-in user [:addresses :registered])
-        context (-> {:recipient (util/user->recipient user)}
+        gid (:google-election-id election)
+        voter-info (voter-info/lookup registered-address gid)
```
cap10morgan: So is this just opportunistically using `gid` in case it's non-nil? Does it do the right thing when it _is_ nil?

src/election_notification_works/voter_info.clj
```diff
@@ -0,0 +1,54 @@
...
+            [turbovote.resource-config :refer [config]]))
+
+(defn google-civic-info-fetch
+  "When given a US `address` string and a `google-election-id` for an
+  election,relevant to that `address`, return the voter information
+  according to the Google Civic Info API."
+  [address google-election-id]
+  (let [api-key (config [:google-api :api-key])]
+    (google-civic/voter-info api-key address google-election-id
+                             {:tries 10 :sleep 1000 :decay :double})))
```
cap10morgan: Should we make `google-election-id` optional?

src/election_notification_works/handlers/vote_tomorrow.clj
zonotope: yes. [`google-civic`]() does the right thing if `gid` is nil, so i sent it through so we can make use of it on the off chance that it's non-nil.

src/election_notification_works/voter_info.clj
zonotope: yeah, that makes sense. this should be a variadic function. i think we'll probably still use the 2-arity version in the existing code paths, but we should also have a version without gid so we don't keep calling it with nil when we know we don't care about the gid.

src/election_notification_works/handlers/vote_tomorrow.clj
zonotope: sorry, i meant to link to [this google civic code](https://github.com/democracyworks/google-civic-clojure/blob/master/src/google_civic/core.clj#L86-L88) in the above comment.

PR #130, opened 2018-03-19
Split upcoming election SMS into two messages 

theJohnnyBrown:
with voter registration status URL


test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -625,21 +625,25 @@
        (let [first-message (:message (take-or-timeout channels/notification-schedule))
              second-message (:message (take-or-timeout channels/notification-schedule))
              third-message (:message (take-or-timeout channels/notification-schedule))
+             fourth-message (:message (take-or-timeout channels/notification-schedule))
```
cap10morgan: Should we update the test description on line 618 to say "publishes four notification..."

theJohnnyBrown: 👍

PR #134, opened 2018-04-13
update absentee ballot request links

zonotope:
pivotal card: https://www.pivotaltracker.com/story/show/156696203

states updated:
ME:
https://tvote.org/2JEecdU

UT:
https://tvote.org/2GRMH2U

ND:
http://tvote.org/1NDsJ8v

VA:
http://tvote.org/2mC7YCD

DC:
https://tvote.org/2HfoMK1


zonotope:
not that i know of. i think we should use https here, but i'll let others weigh in before making that change. i'm just going to merge this now, but we can revisit using https for all the links in another pr.

@cap10morgan @tie-rack thoughts on https?

mrichards42:
See also #136 if there are concerns about https

PR #137, opened 2018-06-06
Format referendums

zonotope:
This patch addresses https://www.pivotaltracker.com/story/show/158089464.

First, I've changed the predicate that detects referendums to take into account differences between the google-civic test election response and the response we got from the production California election. I tried to make the predicate as lenient as possible, but there might be other issues with this if the response returned by the google civic api changes in the future

Secondly, this adds an `:else` clause so data is still returned when there's an unrecognized contest in the google-civic response.


src/election_notification_works/voter_info.clj
```diff
@@ -42,7 +46,8 @@
                                             conj (format-referendum contest))
 
               (race? contest) (update info-map :races
-                                      conj (format-race contest))))
+                                      conj (format-race contest))
+              :else info-map))
```
theJohnnyBrown: What would it mean if we got a contest that didn't look like a referendum or a race? We might want to know about that through an error message instead of passing it along

ddellacosta:
@zonotope Sorry, somehow I missed this one! FWIW LGTM 👍

PR #141, opened 2018-08-02
small change on Spanish dates

li-at-dw:
as part of the work on https://www.pivotaltracker.com/n/projects/930636/stories/158660802


li-at-dw:
I re-ran the build, and now it passed.

li-at-dw:
I checked with Esha, and there's no need to regenerate the notys.

katrinamariehh:
LGTM 🦋

PR #142, opened 2018-08-03
Remove extra upcoming-election/vote-tomorrow notys

mrichards42:
Pivotal cards: [upcoming-election](https://www.pivotaltracker.com/story/show/159406637) and [vote-tomorrow](https://www.pivotaltracker.com/story/show/159406640)

This PR requires [notification-works/#56](https://github.com/democracyworks/notification-works/pull/56) and [notification-works/#57](https://github.com/democracyworks/notification-works/pull/57) to be merged first.

Now that multi-part sms messages can be sent using a single notification scheduled message, we can remove the code that schedules multiple separate notifications (in upcoming-election and vote-tomorrow handlers).


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -126,31 +126,14 @@
...
-        :context context}
-       {:send-to "election-notification-works.vote-tomorrow.update"
-        :send-after send-after
-        :identifiers {:recipient-id recipient-id
-                      :election-id election-id}
-        :notification-type :address-notification-and-change
-        :medium medium
-        :language language
-        :context context}])))
+    {:send-to "election-notification-works.vote-tomorrow.update"
```
theJohnnyBrown: Will we need to do anything different under `handle-update` or is that only changing the context?

mrichards42: Right, `handle-update` only updates the context, and the notys were being sent to `handle-update` individually already, so I don't think there need to be code changes in the update.

PR #146, opened 2018-09-14
Delete and restart notifications by phone number

eggsyntax:
https://www.pivotaltracker.com/story/show/159638701

Add `delete-by-phone` and `restart-by-phone` services, in support of being able to handle stop/start sms messages internally, from `turbovote-http-api` (see the [equivalent branch there](github.com/democracyworks/turbovote-http-api/compare/handle-sms-responses)).


eggsyntax:
Note that commit failures were because I hadn't linked all my github emails to buildkite yet.

resources/config.edn
```diff
@@ -18,6 +18,16 @@
...
+    {:queue "election-notification-works.subscription.delete-by-phone"
+     :f election-notification-works.handlers/delete-subscription-by-phone
+     :response true
+     :prefetch-limit #resource-config/edn #resource-config/env "SUBSCRIPTION_DELETE_PREFETCH"
+     :threads #resource-config/edn #resource-config/env "SUBSCRIPTION_DELETE_THREADS"}
+    {:queue "election-notification-works.subscription.restart-by-phone"
+     :f election-notification-works.handlers/restart-subscription-by-phone
+     :response true
+     :prefetch-limit #resource-config/edn #resource-config/env "SUBSCRIPTION_DELETE_PREFETCH"
+     :threads #resource-config/edn #resource-config/env "SUBSCRIPTION_CREATE_THREADS"}
```
mrichards42: These should be new env vars rather than re-using existing ones.

src/election_notification_works/handlers.clj
```diff
@@ -82,6 +81,34 @@
...
+      {:status :error
+       :error {:type :semantic
+               :message (str "Subscription-delete successes: " (:ok grouped)
+                             "Subscription-delete failures: " (:error grouped))}}
+      {:status :ok
+       :message (str "Deleted " (count (:ok grouped)) " subscription(s).")})))
+
+(defn delete-subscription-by-phone
+  "Given a map with a phone number, delete all sms subscriptions associated with this phone number."
+  [message]
```
mrichards42: I won't hold the PR up for this, but I would if this were `delete-subscription-by-attribute` or `delete-subscription-by-search` or something like that. And then either of these messages could be passed:
```clj
{:phone "1234567890"}
{:email "hello@world.com"}
```

That would also mean renaming `user/users-by-phone` to `user/search-users` or something like that . . . and I'll resume this thread in a comment on that function.

src/election_notification_works/user.clj
```diff
@@ -19,3 +22,17 @@
     (do
       (log/error "Timeout while reading user for id: " user-id)
       nil)))
+
+(defn-traced users-by-phone
+  "Given a phone-number string, returns all users matching that phone number."
+  [user-phone]
+  (if-let [response (async/<!! (search-user-by-phone {:search-phone user-phone
+                                                      :hostnames :all}))]
```
mrichards42: Picking up where we left off in `delete-subscription-by-phone`, this could be something along these lines:
```clj
(defn search-users [message]
  (let [search-message (rename-keys message {:search :search-phone, :email :search-email})
        response (async/<!! (search-user-fn (assoc message :hostnames :all)))]
    ...))
```

src/election_notification_works/handlers.clj
```diff
@@ -95,6 +122,34 @@
...
+(defn- sms-subscribed [user]
+  (-> (read-subscription {:user-id (:id user)}) :subscription :mediums :sms))
+
+(defn- latest-unsubbed-account-for-phone
+  "Given a phone number, return the most recently created account with that phone
+  number which is not currently sms-subscribed."
+  [user-phone]
+  (let [accounts (user/users-by-phone user-phone)]
+    (first (sort-by :created-at
+                    (remove sms-subscribed accounts)))))
```
mrichards42: If there are already subscribed users with this phone number, should `START` be a no-op?

src/election_notification_works/handlers.clj
```diff
@@ -82,6 +81,34 @@
...
+  (let [user-phone (:phone message)
+        users (user/users-by-phone user-phone)
+        user->sms-subscription (fn [u] {:user-id (:id u), :mediums #{:sms}})
+        delete-responses (for [user users]
+                           (delete-subscription (user->sms-subscription user)))]
+    (if (seq users)
+      (collate-delete-responses delete-responses)
+      {:status :error
+       :error {:type :semantic
+               :message (str "No users found with phone number " user-phone)}})))
```
mrichards42: It looks like we use [this pattern](https://github.com/democracyworks/user-works/blob/5e8b68fce818c9ff286fe7d27c1953dfa08f2f4c/src/user_works/handlers.clj#L343) for user-not-found sort of errors (also found in [`election-mail-works`](https://github.com/democracyworks/election-mail-works/blob/9854f11365a7468f0a60a3790ce0258215387738/src/election_mail_works/handlers/subscription.clj#L27))

resources/config.edn
eggsyntax: Whoops, meant for those both to be CREATE, not one as DELETE. It seemed to me that this was just a particular case of CREATE, but I'll add new ones. I see that the new ones'll need to be added to the service template. But I assume that means that those will need to be added somewhere else entirely? It looks like the values are being pulled from somewhere via `curl -s http://${COREOS_PRIVATE_IPV4...`, but where's that?

src/election_notification_works/handlers.clj
mrichards42: I'm not sure I follow. `user-works` exposes a single search queue: `user-works.user.search` (well, also `user-works.user.search.streaming`, but that's about the response type).

src/election_notification_works/handlers.clj
eggsyntax: IIRC, I landed on this behavior based on a conversation with @rackette, based on the possibility that more than one user might be subscribed using the same number. I'll see what she thinks.

rackette: @mrichards42 @eggsyntax This also matches the behavior the team agreed on when we pointed this card and is reflected in the COAs. https://www.pivotaltracker.com/story/show/159638701

src/election_notification_works/handlers.clj
eggsyntax: That reminds me, you had tossed out the idea of inheriting the most severe of the individual errors, but it wasn't clear to me that we actually had a useful hierarchy there. Thoughts?

resources/config.edn
mrichards42: I think keeping them separate is a good plan, if just to make it consistent. Actually, based on the [buildkite pipeline](https://github.com/democracyworks/election-notification-works/blob/deb0f092666eb22e7932b76c63e55c589ba5dd80/.buildkite/pipeline.yml) this is a tv-only service (i.e. not shared w/ ballot scout), so that at least makes the deploy a little easier. That said, the fleet service template (and consul, which is where it was pulling env vars from) is no longer in use. To actually get this deployed, we need to merge the `ro-deploy` branch into master first, then you can add the env vars to https://github.com/democracyworks/election-notification-works/blob/98529c288e48faebe7104058aae6975c61e9a155/deploy/development/election-notification-works.configmap.yml

mrichards42: Looks like here as well: https://github.com/democracyworks/election-notification-works/blob/98529c288e48faebe7104058aae6975c61e9a155/deploy/production/election-notification-works.configmap.yml

src/election_notification_works/handlers.clj
eggsyntax: You're 100% right; I either misremembered or drew the wrong conclusion from the implementing code in `user-works`. That said, since it seems like a priority to get this out by 9/18, I propose we skip it for now and then I'm happy to make that my next task once the initial version is in production.

src/election_notification_works/handlers.clj
eggsyntax: @rackette could you clarify "this" in your comment? Do you mean the current behavior as currently implemented reflects the COAs, or Mike's suggestion does?

src/election_notification_works/handlers.clj
mrichards42: Don't worry about it. There's a compelling enough case that `delete-by-phone` and `delete-by-email` are sufficiently different enough operations to warrant different queues. And I'm not even sold on `delete-by-search` being the right thing anyways (e,g, `delete-by-fulltext-email-search` is obviously wrong, as are `delete-by-hostname-search`, etc).

Plus this code will be really easy to refactor to add an email queue, so we can do that when we need it.

src/election_notification_works/handlers.clj
```diff
@@ -82,6 +81,34 @@
...
+  (let [user-phone (:phone message)
+        users (user/users-by-phone user-phone)
+        user->sms-subscription (fn [u] {:user-id (:id u), :mediums #{:sms}})
+        delete-responses (for [user users]
+                           (delete-subscription (user->sms-subscription user)))]
+    (if (seq users)
+      (collate-delete-responses delete-responses)
+      {:status :error
+       :error {:type :not-found
+               :message (str "No users found with phone number " user-phone)}})))
```
mrichards42: I'm not 100% sure what the response should be for this case, but `delete-subscription` returns `{:status :ok}` for "user not found" and for "user found, but already unsubscribed". I'm leaning toward this should also return `{:status :ok}` for user not found.

src/election_notification_works/handlers.clj
```diff
@@ -6,18 +6,17 @@
...
             [election-notification-works.handlers.util :as util]
             [clojure.tools.logging :as log]
             [schema.core :as s]
             [datomic-toolbox.core :as db]
-            [election-notification-works.notification :as notification]
-            [election-notification-works.handlers.sms-form :as sms-form]
             [clojure.core.async :as async]
             [clojure.set :as set]
             [kehaar.wire-up :as wire-up]
-            [election-notification-works.notification :as notification]
```
theJohnnyBrown: Did you mean to remove both of the dupe requires?

src/election_notification_works/handlers.clj
theJohnnyBrown: @eggsyntax the current behavior, COA calls for re-subscribing the account with the most recent :created-at AND a value of sms-subscribed = FALSE

eggsyntax: Cool, that was my impression (that behavior as implemented in PR matches COA), just wanted to be sure.

src/election_notification_works/handlers.clj
eggsyntax: I did, it turned out to be unused.

src/election_notification_works/handlers.clj
```diff
@@ -82,6 +81,37 @@
   (errors/handle-errors (partial delete-subscription*
                                  channels/subscription-delete-events-out)))
 
+(defn delete-subscription-by-phone*
+  ([message]
+   (delete-subscription-by-phone* user/users-by-phone delete-subscription subscription/read-subscriptions message))
+  ([users-by-phone-fn delete-subscription-fn read-subscription-fn message] ; for testing
+   (log/trace "Deleting sms notys for " (:phone message))
+   (let [user-phone (:phone message)
+         users (some-> user-phone users-by-phone-fn)
```
mrichards42: Rather than allowing a nil phone number, this and `restart-subscription-by-phone` should have a spec that's validated at the start of the function (similar to `delete-subscription` and `upsert-subscription`). A lot of our projects have moved to `clojure/spec.alpha`, but it looks like this one's still using `prismatic/schema`. Sorry, I must have glossed over that on my initial read through.

src/election_notification_works/handlers.clj
```diff
@@ -82,6 +81,37 @@
...
+         user->sms-subscription (fn [u] {:user-id (:id u), :mediums #{:sms}})
+         delete-responses (for [user users]
+                            (delete-subscription-fn (user->sms-subscription user)))
```
mrichards42: How about `assoc`-ing the user ids with the return from `delete-subscription-fn` so the caller knows which users succeeded and which failed?

src/election_notification_works/handlers.clj
eggsyntax: Gotcha. Glad to hear we've been moving to spec, I've liked it WAY better than schema once I got used to it.

src/election_notification_works/handlers.clj
mrichards42: Since the work of this handler is calling `delete-subscription` for each user I do think it would make sense for the return to include relevant return data from that function. Also, from what I can tell, `delete-subscription` already returns the updated subscription, so I don't think you need to fetch them again. I didn't notice until now that `subscription/read-subscription` already includes the user ids, so you probably don't need to add them as I suggested.

mrichards42: So, ultimately I think the return would be best as one of:
```clj
{:status :ok
 ;; either
 :subscriptions [{:user-id ... :mediums #{}}, ...]
 ;; or
 :subscriptions {<userid> <mediums>, ...}
}

{:status :error
 :error {:type :semantic}
 ;; either on the root, or inside the :error map
 :subscriptions [...] ;; or {} as above, no real preference
}

eggsyntax: No message if `:ok`, then? Or was that an omission?

eggsyntax: I figured it couldn't hurt to bundle it in, even if the caller may choose to ignore it.

mrichards42: I haven't come across a `{:status :ok}` response that include a `:message` so that was intentional (that doesn't mean there isn't one, but I don't think it's common if so). If there's a message you want somebody to see, I'd `log/debug` it here.

And actually, I'd `log/error` a message if there's an error. Usually that would happen in `handle-errors`, but since this error isn't actually being thrown it wouldn't get logged otherwise.

eggsyntax: Hmm, actually as I look at it, `delete-subscription` doesn't consistently return subscriptions -- it does in the most common case, but doesn't if a) `handle-errors` catches something, b) `delete-subscription` is a no-op (ie the user already isn't subscribed by sms).

That leaves me inclined to continue to use `read-subscription` to get the updated list of subscriptions to put in the return. Thoughts?

eggsyntax: I didn't log the error because it looked like it would get logged via the `handle/errors` wrapper on `delete-subscription`.

mrichards42: I guess I'm inclined to say "do whatever `delete-subscription` does" which is to say . . .

a) If an exception was thrown, the caller doesn't expect to get anything useful back, aside from an error message. If they need to retry, they can retry.

b) In this case we should return the subscription -- `delete-subscription` should probably be changed to return the `before` subscription.

All that said, I wasn't able to find a caller in any of our services that actually uses the return value for `delete-subscription`, so this might be purely academic anyways.

PR #149, opened 2018-09-27
upcoming-election deadline context 

li-at-dw:
https://www.pivotaltracker.com/n/projects/930636/stories/160797226

To add OVR deadlines for SMS, we make the deadline context the same for SMS and email.


katrinamariehh:
LGTM 🌊

PR #150, opened 2018-10-09
don't notify users who've told us they can't vote

theJohnnyBrown:


src/election_notification_works/registration_status.clj
```diff
@@ -0,0 +1,19 @@
...
+
+
+(defn not-ineligible?
+  "This is used to pre-empt notifications, so we return true unles a user has
+  positively identified as ineligible"
+  [user-id]
+  (let [resp (clojure.core.async/<!! (reg-status-read [user-id]))]
+    (case (:status resp)
+      :ok (not= (-> resp :registration-status :status) :ineligible)
+      :not-found true
```
eggsyntax: In what sorts of cases would `:status` be `:not-found`?

test/election_notification_works/handlers/signup_test.clj
```diff
@@ -226,100 +232,61 @@
...
-              expected [signup-result confirm-result]
-              results (atom [])]
-          (reschedule (success-search-fn (list {:message signup-message}
-                                               {:message confirm-message}))
-                      (fn [m] (swap! results conj m) (async/go {:status :ok}))
-                      user-id)
-          (is (= expected @results))))))
-
-  ;; The following tests check that signup events trigger a reschedule
-  (letfn [(test-event [event-fn]
```
eggsyntax: Out of curiosity, why get rid of these tests? Would they somehow no longer be relevant? I feel like I'm missing something.

src/election_notification_works/registration_status.clj
theJohnnyBrown: In cases where `voter-registration-works` has no record for the given user ID. For example, users who haven't reached that step of their signup flow will be `not-found`.

test/election_notification_works/handlers/signup_test.clj
theJohnnyBrown: No, I didn't mean to do that. It should be just indentation changes from wrapping things in `with-redefs`. I'll take a closer look.

eggsyntax: Oh, could definitely be github messing up the diff. I told it to ignore whitespace, and I've certainly seen that go wrong occasionally...

src/election_notification_works/registration_status.clj
eggsyntax: Makes sense, thanks!

katrinamariehh:
💅🏻lgtm

PR #151, opened 2018-10-12
Polling hours

li-at-dw:
tweaks on the path to adding polling hours to vote-tomorrow noty.


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -56,19 +56,21 @@
 (defn assoc-polling-places-if-exist
   "Given a context map, and a map of voter info, associates the polling
   place information from the voter info map with the given context
-  map. If none is found, the context map is returned as-is."
+  map. If none is found, the context map is returned as-is.
+  We only return up to 3 polling places."
   [context voter-info]
   (if-let [polling-places (:pollingLocations voter-info)]
-    (assoc context :polling-places polling-places)
+    (assoc context :polling-places (take 3 polling-places))
```
theJohnnyBrown: Is there any kind of sort that would make sense to apply here, for the 3 "best" polling places? Or maybe the relevant information isn't available

li-at-dw: The API returns the polling places sorted by proximity to the users' address, and we only want to show the top 3 in the noty.

PR #152, opened 2018-10-16
check voter eligibility when sending, as they probably signed up

theJohnnyBrown:
before telling us their eligibility


src/election_notification_works/handlers/signup.clj
```diff
@@ -87,7 +87,7 @@
                                (sub/read-subscription (dt/db))
                                :mediums
                                (into #{} filter-valid))]
-    (when (seq subscriptions)
+    (when (and (seq subscriptions) (not-ineligible? user-id))
```
mrichards42: How come this is `not-ineligible?` instead of `eligible?`

li-at-dw:
sorry, I just now realized that `reschedule` calls `schedule`, and that was the entire point of this PR.

li-at-dw:
I'm still confused....if the message was already scheduled, and then the user becomes ineligible, where  are we actually unscheduling the message?

theJohnnyBrown:
@mrichards42 and @li-at-dw, this PR might be a little confusing because it's not entirely standalone, it addresses some issues with [the previous PR](https://github.com/democracyworks/election-notification-works/pull/150). Vote-tomorrow and upcoming-election are both handled there.

@li-at-dw We don't unschedule, we just check again when the "send" message comes back here from scheduled messages works. The reason for for the delay in sending initial notifications is so we can wait a few minutes and check for any additional data the user entered before sending.

li-at-dw:
@theJohnnyBrown Thanks...makes much more sense now...

theJohnnyBrown:
Yeah, it would make sense to add a check to `handle-update` for vote-tomorrow and upcoming-election as well

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -171,21 +172,22 @@
...
-                        (assoc-polling-place-urls-if-exists election)
-                        (assoc-polling-places-if-exist voter-info)
-                        (assoc-drop-off-locations-if-exist voter-info)
-                        (assoc-ballot-info-if-exists voter-info)
-                        update-authority-data))
-        updated-notification (update notification :context update-fn)]
-    (notification/send updated-notification)
-    {:status :ok
-     :notification updated-notification}))
+  (when (not-ineligible? (:id recipient))
```
mrichards42: My understanding is that we're trying to avoid as much rpc as we can at send time. Would it work to delete notys in the `electorate.create` handler when we discover a user is ineligible? Afaik those events are sent for each user update.

src/election_notification_works/registration_status.clj
```diff
@@ -5,9 +5,6 @@
 (def reg-status-read
   (wire-up/async->fn channels/reg-status-stream-read))
 
-(def user-id #uuid "5b870209-8fef-4cb1-bb3f-46c6ddeac28b")
-
-
 (defn not-ineligible?
```
mrichards42: Do we care about checking user `:dob` and `:citizenship` as well? Or do we update the reg status to be ineligible when we find out that a user is < 18 or not a us citizen?

PR #154, opened 2018-10-24
use DOB to determine voter eligibility when present, unschedule notys…

theJohnnyBrown:
… when user updates to ineligible


src/election_notification_works/handlers/user.clj
```diff
@@ -0,0 +1,25 @@
...
+;; any of the below events will trigger a potential rescheduling
+
+(defn-traced event-user-update*
+  [search-fn schedule-fn {:keys [after] :as msg}]
+  (let [user-is-notifiable (reg/not-ineligible? (:id after))]
+   (if user-is-notifiable
+     (signup/event-user-update msg)
+     (do
+       (log/debug "unscheduling " after)
+       (util/unschedule-notification {:recipient-id (:id after)})))))
```
mrichards42: Could this be done in the `electorate-create` event handler, since there's already an eligibility check in that code? From what I can tell, a user update triggers an electorate create. Or if you'd rather use a more specific event, I think the `registration-status.create` event would be more accurate than user update.

src/election_notification_works/handlers.clj
```diff
@@ -248,7 +248,9 @@
           (handler electorate valid-mediums)
           (catch Exception e
             (log/error e
-                       "Exception calling a notification-type handler")))))))
+                       "Exception calling a notification-type handler")))))
+    (when (not eligible-voter?)
+      (util/unschedule-notification {:recipient-id (:id user)}))))
```
mrichards42: I'm ok w/ this delete all notys for the user, unlike `handle-electorate-delete*`, which deletes just notys for this user/election. But I think it would be good to limit it to `{:notification-type :vote-tomorrow}` and `:upcomng-election` as well, in case we ever need to schedule a non-noty message with `:recipient-id` as an identifier.

theJohnnyBrown:
@mrichards42 anything else you want to see here?

PR #155, opened 2018-10-24
Query civic info API at schedule time too & cache result when allowed

cap10morgan:
Let me know if this wasn't what you had in mind.


mrichards42:
While I'm thinking about it, this is what I'm considering for the whitelist/blacklist approach. I'll explain my thinking and see if it makes sense to you.

I'm thinking we have both a whitelist (hopefully state-level ocd-ids) and a blacklist (hopefully county-level or smaller). So the process of determining which notys can skip requesting voter-info might look like:

```clj
(defn can-skip-voter-info? [context]
  (and (some-voter-info-key-is-present? context)
       (recipient-ocd-id-in-whitelist? context)
       (not (recipient-ocd-id-in-blacklist? context)))
```

Having to check the timestamp makes this a little more complicated. For the whitelist, maybe it's filter the whitelist by recipient ocd-id and make sure the `:voter-info-timestamp` is after all of the whitelist timestamps.

I'm not sure what the semantics of timestamps in a blacklist would be . . . so maybe the blacklist doesn't use the timestamp.

cap10morgan:
@mrichards42 Ready for re-review now. Good suggestions / catches. I added some tests too.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -115,10 +112,53 @@
   [{:keys [election district-division] :as electorate}]
   (let [voting-id-instructions (some->> (:voting-methods district-division)
                                         (filter vm/in-person?)
-                                        (some #(get-in % [:instructions :voting-id])))]
-    (-> (select-keys election [:description :google-election-id])
+                                        (some #(get-in % [:instructions
+                                                          :voting-id])))]
+    (-> (select-keys election [:description :google-election-id
+                               :polling-place-url :polling-place-url-shortened])
```
mrichards42: Do we need to include the two polling-place keys since they're being assoc'ed from the election in `update-voter-info`?

mrichards42: Oh, I follow. The old code was calling `(assoc-polling-place-urls-if-exists election)` in the [update handler](https://github.com/democracyworks/election-notification-works/blob/master/src/election_notification_works/handlers/vote_tomorrow.clj#L182), when all it had available was the election from the context. So that must have been an unintentional no-op, with the result that polling place urls were only `assoc`ed from the state authority.

cap10morgan: Yeah, the polling-place-url tests fail when I comment those out.

mrichards42: 👍 carry on

cap10morgan:
I think I'm going to hold off on merging this until the allow/deny list aspect is ready too.

cap10morgan:
@mrichards42 I added your allow & deny list logic in the 2 most recent commits. Let me know what you think.

resources/config.edn
```diff
@@ -278,4 +278,7 @@
                  "tuamericacorreo.turbovote.org"
                  "myfaithvotes.turbovote.org"
                  "becauseicare.turbovote.org"}
-               :user-batch-size 100}}
+               :user-batch-size 100}
+ :voter-info-cache
+ {:allow-list {"ocd-division/country:us/state:co" {:valid-as-of #inst "2018-10-25"}}
+  :deny-list  #{"ocd-division/country:us/state:co/county:jefferson"}}}
```
cap10morgan: These are here as placeholder examples. We should probably get the real data in before merging.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -115,10 +112,83 @@
...
         (assoc-in [:instructions :voting-id] voting-id-instructions))))
 
+(def voter-info-keys
+  #{:voter-info-timestamp :polling-place-url :polling-place-url-shortened
+    :polling-places :drop-off-locations :ballot-info})
```
mrichards42: I think we only want keys that are added by the `voter-info/lookup` query in this list, but not the ones that are added from the election or election authority, since those aren't dependent on VIP data. I believe the election + authority keys would usually be present, so `(some voter-info-keys (keys context))` would typically be true, even if the voter-info lookup actually failed to return data.

Also from what I can tell we're relying on updating id instructions from the state authority at send time so that we can accommodate last-minute changes to that language.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -115,10 +112,83 @@
...
+  (let [allow-list (or allowed (config [:voter-info-cache :allow-list] {}))
+        cache-timestamp (some-> context :voter-info-timestamp coerce/from-date)]
+    (boolean
+     (when cache-timestamp ; time/before? returns true when 2nd arg is nil for some reason
+       (some-> context
+               (get-in [:recipient :addresses :registered :ocd-divisions])
+               (->> (some #(get allow-list %)))
+               :valid-as-of
+               coerce/from-date
+               (time/before? cache-timestamp))))))
```
mrichards42: I was thinking it could be possible to have multiple ocd-ids match the allow list, and we would require that all of the `:valid-as-of` timestamps be earlier than the `:voter-info-timestamp`.

So maybe we start with

```clj
{:allow-list {"ocd-division/country:us/state:co" {:valid-as-of #inst "2018-10-25"}}}
```

Then we learn that just jefferson county had to be updated (but the rest of CO was ok), so we add

```clj
{:allow-list {"ocd-division/country:us/state:co" {:valid-as-of #inst "2018-10-25"}
              "ocd-division/country:us/state:co/county:jefferson" {:valid-as-of #inst "2018-10-30T15:00:00"}}}
```

Then again, we might be able to get rid of the deny list with this approach, so I'm open to options

cap10morgan: This sounds like something to have Esha and Maria weigh in on, perhaps?

src/election_notification_works/handlers/vote_tomorrow.clj
cap10morgan: OK, so it sounds like the fix we need here is to remove the `:polling-place-url` and `:polling-place-url-shortened`. But let me know if I'm missing the implications of anything else here.

mrichards42: I think the other part is that this part of `update-voter-info` should be called regardless of whether or not we're fetching new voter info:
```clj
(-> context
    (assoc-polling-place-urls-if-exists election)
    update-authority-data)
```

cap10morgan: Ohhhh, gotcha. Yes, good catch.

cap10morgan: OK, see what you think of that. It seemed like it was time to separate the two types of voter-info.

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -115,10 +112,93 @@
...
+  notification `context`."
+  [{:keys [election] :as context}]
+  (-> context
+      (assoc-polling-place-urls-if-exists election)
+      update-authority-data))
+
+(defn update-voter-info [context]
+  (-> context
+      update-voter-info-from-vip
+      update-vote-info-from-election-authority))
```
mrichards42: Typo on `vote` instead of `voter`? But otherwise splitting that into 2 functions makes a lot of sense to me

cap10morgan: Yep, but at least I was consistent! 😆

src/election_notification_works/handlers/vote_tomorrow.clj
mrichards42: After talking w/ Esha, Maria, and Katy, it sounds like we'll be able to get correction data from VIP at the county level. I think the above plan accounts for the following possible issues:

1. We update a whole state on 11/1
2. Several counties in that state need to change their data on 11/2 (so we re-run just those counties)
3. Another county in the same state is found to have bad data on morning of the day we send voter-tomorrows (we could either place that county on the `deny-list` or add it to the `allow-list` with a date in the future).

cap10morgan: OK, so it sounds like you want me to match every OCD-ID in the allow-list and ensure that the data timestamp is after every `valid-as-of` date or re-query the API?

I think I'm in favor of keeping the deny-list for scenarios where we don't when / if there will be valid data in the future (rather than adding a potential time bomb to the system).

cap10morgan:
I think this is ready for final review, @mrichards42. Want me to remove the placeholder config values and leave the `:allow-list` and `:deny-list` empty for now?

mrichards42:
👍 LGTM thanks for taking this on!

Yes, let's remove the placeholder config and leave those keys empty for now.

PR #158, opened 2018-11-01
Test that voter-info-cache is formatted correctly

mrichards42:
I screwed up when adding values to `:allow-list` earlier today, and discovered that it can prevent messages from being sent (by throwing an exception in `vip-voter-info-cache-valid?`)

I'm not sure if this is the best way to do this, so I'm very open to feedback. In particular, the `(config-file "../resources/config.edn")` bit feels pretty hacky, but since there's already a `test-resources/config.edn` file, I couldn't find another way to reference the prod config.


mrichards42:
1. Looks like we're using it in tests -- lots of them fail if I remove it.
2. This is probably not the best way to do it, but it seems reasonable to me to have some kind of automated test that your config is formatted correctly before it gets deployed. Now that I think about it more, `spec` is a much better fit than what I have here.
3. With or without this PR, sticking a comment in config.edn is a good plan

PR #159, opened 2018-11-02
IL, NY, NJ, NV, TN, and ME are good

tie-rack:


mrichards42:
Let's make those `:valid-as-of #inst "2018-11-02"` so that any users created between 10-30 and 11-02 will re-fetch the data

mrichards42:
👍 still looking good

PR #164, opened 2018-11-09
Move default google-election-id into config

li-at-dw:


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -186,8 +186,7 @@
   (if (vip-voter-info-cache-valid? context)
     context
     (let [registered-addr (get-in recipient [:addresses :registered])
-          ; 6000 is the suggested default election ID for the 2018 Primary election
-          gid (or (:google-election-id election) 6000)
+          gid (or (:google-election-id election) (config [:google-api :default-election-id]))
```
mrichards42: let's throw a `nil` default value on the end of that just in case we forget it in the config (since `config` throws without a default)

PR #166, opened 2018-12-07
Delete users by email address

eggsyntax:
Adds a service for deleting subscriptions by email address; does some generalization and adds more tests.

Should be merged after [krak PR 76](https://github.com/democracyworks/krakenstein/pull/76), which adds the relevant environment vars.


src/election_notification_works/handlers.clj
```diff
@@ -84,37 +84,85 @@
...
+
+(defn delete-subscriptions-by-emails*
+  ([message]
+   (delete-subscriptions-by-emails* user/users-by-email delete-subscription message))
+  ([users-by-email-fn delete-subscription-fn message] ; for testing
+   (s/validate by-emails-schema message)
+   (log/debug "Deleting email notys for" (:emails message))
+   (let [user-emails (:emails message)
+         users (mapcat users-by-email-fn user-emails)]
+     (delete-subscriptions delete-subscription-fn users :email))))
```
mrichards42: It seems like overkill to me to have separate queues for deleting email subscriptions, especially since the delete-with-multiple-emails queue fires off a message per-user here anyways. Are there compelling use cases for deleting with a single email and deleting with multiple emails? If there is a compelling case for both, I wonder if it would be simpler to have a single function that takes either `:email` or `:emails`.

test/election_notification_works/handlers_test.clj
```diff
@@ -362,28 +385,62 @@
...
+          user (test-user uid)
+          email-map (select-keys user [:email])
+          users-by-email-fn (constantly ())
+          delete-response {:status :ok}
+          delete-subscription-fn (constantly delete-response)]
+      (is (= {:status :ok, :message "No matching users found"}
+             (delete-subscription-by-email* users-by-email-fn
+                                            delete-subscription-fn
+                                            email-map))))))
+
```
mrichards42: Now that I'm reading through these tests a second time (the first being the delete-by-phone review), it would be good to have a couple of tests that use the actual delete functions instead of mocking everything. I think as a rule we always mock external services in tests, but since this is effectively a wrapper around an internal service (`delete-subscription`) let's also make sure that the correct subscriptions are actually deleted. I'm thinking of something along the lines of [this test](https://github.com/democracyworks/election-notification-works/blob/8ff17b8face51a679b5f405b93aeaeed34abc5b9/test/election_notification_works/handlers_test.clj#L186), which creates an actual subscription in the test db and then deletes it.

src/election_notification_works/user.clj
```diff
@@ -23,16 +23,27 @@
...
+  (if-let [response (async/<!! (search-user search-map))]
     (if (= :ok (:status response))
       (:users response)
       (do
-        (log/error "Error reading user " (pr-str response))
+        (log/error "Error reading user:" (pr-str response))
         nil))
     (do
-      (log/error "Timeout while reading user for phone: " user-phone)
+      (log/error "Timeout while reading user: " search-map)
```
mrichards42: Couple minor quibbles:
1. can we `pr-str` the search map?
2. Should that be "reading user" or "searching for user"?

src/election_notification_works/handlers.clj
```diff
@@ -84,37 +84,85 @@
...
+  ([delete-subscription-fn users medium] ; for testing
+   (let [user->subscription (fn [u] {:user-id (:id u), :mediums #{medium}})
          delete-responses (for [user users]
-                            (delete-subscription-fn (user->sms-subscription user)))
+                            (delete-subscription-fn (user->subscription user)))
          grouped-responses (group-by :status delete-responses)]
      (cond
        (empty? users) {:status :ok
-                       :message (str "No users found with phone number " user-phone)}
+                       :message "No matching users found"}
```
mrichards42: It might be more consistent to have this case return an empty map for the `:subscriptions` key? (or a nil for that key?)

src/election_notification_works/handlers.clj
eggsyntax: No, it seems perfectly reasonable to me to eliminate the singular version -- things are a bit blurry on that since I started working on this before election & onsite, but IIRC the singular version was created first and then later when I understood the SNS notification format better, it became clear that I needed the multiple version. But it's easy enough to call `(delete-subscriptions-by-email delete-fn ["just@one.email"])`. I'll get rid of the singular version.

mrichards42: I think I would prefer that the delete-by-phone and delete-by-email queues both take either one or multiple, and I'd argue that most of our endpoints do single operations rather than a batch, so (though we're already in fuzzy territory here since these are effectively batch operations, whether or not they take a single email/phone or multiple emails/phones)

eggsyntax: I'm not necessarily opposed to that, but my initial instinct is to be a bit resistant to it. I'm not sure why, trying to think it through. One downside of that is that it pushes more logic into `turbovote-http-api` (mapping the service calls across multiple emails, but also deciding what it means to create a collated response from multiple service calls), which already has to deal with a lot of logic for dealing with SNS calls, and I know we've wanted to keep `t-h-a` dumb where we can.

src/election_notification_works/handlers.clj
eggsyntax: Good thought.

src/election_notification_works/user.clj
eggsyntax: 1. Yep, still haven't quite gotten in the habit of routinely `pr-str`ing.
2. Whups, copypasta error.

src/election_notification_works/handlers.clj
mrichards42: I'm more concerned w/ them being consistent, whether that means they both take multiple, or both take single, or both take either multiple or single.

mrichards42: and absolutely👍 for trying to keep t-h-a as dumb as possible

test/election_notification_works/handlers_test.clj
```diff
@@ -338,21 +338,63 @@
...
+          delete-response {:status :ok, :subscription nil}
+          _ (upsert-subscription* (async/chan)
+                                  {:user-id uid
+                                   :mediums #{:sms}
+                                   :suppress-confirmation true})
+          delete-result (delete-subscription-by-phone* users-by-phone-fn
+                                                         delete-subscription
+                                                         phone-map)]
+      ;; Consume the deletion broadcast event so other tests aren't confused
+      (async/<!! channels/subscription-delete-events-out)
```
mrichards42: It looks like other tests get around this problem by using a dummy channel for events, like
```clj
(let [delete-fn (handle-errors (partial delete-subscription* (async/chan)))]
  (delete-subscription-by-phone* users-by-phone-fn delete-fn phone-map))
```
I'm not sure if this is more or less confusing, but it would serve to further isolate this test from the others

eggsyntax: Good call, I'd forgotten that they take an event-chan

src/election_notification_works/user.clj
```diff
@@ -23,16 +23,27 @@
       (log/error "Timeout while reading user for id: " user-id)
       nil)))
 
-(defn-traced users-by-phone
-  "Given a phone-number string, returns all users matching that phone number."
-  [user-phone]
-  (if-let [response (async/<!! (search-user-by-phone {:search-phone user-phone
-                                                      :hostnames :all}))]
+(defn-traced user-search
```
mrichards42: I think we would usually use `defn-traced` for `user-search` or `users-by-phone` and `users-by-email`, but not all three (as in, either the internal functions or the external functions, but not both)

eggsyntax: Do you have a sense of what's the more common convention?

mrichards42: Reading through what's already in this file, I think what we've been doing is using `defn-traced` for the functions that do the actual work, and not tracing the wrapper functions.

PR #167, opened 2019-02-01
Expose a REPL when running in krakenstein

eggsyntax:
This PR makes the minimal changes necessary to expose an nrepl server when running in krakenstein, and can serve as a model for other services to do the same. See below for instructions on connecting. Not actually exposed from krak until we merge the relevant [krak PR](https://github.com/democracyworks/krakenstein/pull/78).


eggsyntax:
Closing -- no longer needed because [this krak PR](https://github.com/democracyworks/krakenstein/pull/80) does the same thing without having to change this or other services.

PR #168, opened 2019-02-12
Remove everything from allow-list and deny-list

mrichards42:
... this is out of date, based on the november election


mrichards42:
I just re-read the voter-info-cache code, and this is definitely part of the problem we ended up with yesterday. I thought that we were checking to see that we have voter-info data already present in the noty (from when it was created), but effectively we're only checking that that we've ever _tried_ to fetch the data (since `:voter-info-timestamp` is one of the keys we look for).

PR #172, opened 2019-02-21
All mail notys

li-at-dw:
This ticket https://www.pivotaltracker.com/n/projects/930636/stories/163200093

Goes with this PR:  https://github.com/democracyworks/notification-works/pull/91


li-at-dw:
hehe....I was just looking at the frontend and the `elections/upcoming` endpoint, and wondering the same thing myself.  I _think_ so?  I could make `electorate-works` do this logic for `election/voting-methods` and add in the `all-mail-level` flag if the primary voting method is `by-mail`.

That would be cleaner.  Thanks!

li-at-dw:
(I guess what I mean to say is that in the voting-methods, if we get `{:type :by-mail :primary true}`, I would make it `{:type :by-mail :primary true :all-mail-level :state}` or something like that.)

mrichards42:
I was actually thinking that electorate-works could figure out the full list of voting-methods for a user/election and set that under the `:district-division` key in the electorate. But maybe you do also need to add a separate flag that indicates _where_ the voting-methods came from.

mrichards42:
This also makes me realize that I don't think the electorate-works data model is documented anywhere, it's definitely not obvious what the difference is between `:electorate :district-divisions` and just `:district-division`

li-at-dw:
Yeah, the electorate-works data model strictly speaking only has the ids for the user, election, and election and registration authorities.  If I add this, I think I would have to change the `electorate->full-electorate` method.

What it looks like to me is that while  the `:election` key has the raw info for the election (including the all-mail override on the election level), it could technically have multiple `district-divisions`, and so the `:district-division` key was created to pull out the single `district-division` that applies to this electorate--maybe for the convenience that the downstream functions would not have to search all the ocds of the user for the matching district-division.

So, I think updating the `:voting-methods` inside `district-division` _seems_ ok...We definitely need to flag that set the all-mail-primary to true, because  our language for the different cases would be slightly different.

mrichards42:
> Yeah, the electorate-works data model strictly speaking only has the ids for the user, election, and election and registration authorities. If I add this, I think I would have to change the `electorate->full-electorate` method.

Good point. I guess I'm thinking more of api documentation, since all the queues that return electorates call `electorate->full-electorate`.

> What it looks like to me is that while the `:election` key has the raw info for the election (including the all-mail override on the election level), it could technically have multiple `district-divisions`, and so the `:district-division` key was created to pull out the single `district-division` that applies to this electorate--maybe for the convenience that the downstream functions would not have to search all the ocds of the user for the matching district-division.

This makes me wonder if we should add a new top-level `:voting-methods` key that would represent the specific voting methods for this electorate (based on the local authority). If this is a statewide election, `:district-division` is going to be the state, not the local authority, so changing the voting methods might be confusing. I'm not sure, just something to think about.

li-at-dw:
> This makes me wonder if we should add a new top-level `:voting-methods` key that would represent the specific voting methods for this electorate (based on the local authority). If this is a statewide election, `:district-division` is going to be the state, not the local authority, so changing the voting methods might be confusing. I'm not sure, just something to think about.

Yeah, I think you are right that it would be confusing.  I mean, it's already confusing. 😄 It kind of makes sense to have a `:voting-methods` key, meaning that these are the voting-methods to this particular electorate.

li-at-dw:
Closing this PR in favor of https://github.com/democracyworks/election-notification-works/pull/180

PR #174, opened 2019-02-27
Add a bit of extra logging on error

eggsyntax:


src/election_notification_works/voter_info.clj
```diff
@@ -79,10 +80,16 @@
 (defn format-dropoff-locations [dropoff-locations]
   (map assoc-map-url dropoff-locations))
 
+(defn- log-on-error [response]
+  (when-not (= 200 (:status response))
+    (log/info "Error response from Google Civic API:" response))
```
mrichards42: Not that it makes a big difference, but maybe this should be log/warn since it indicates a problem?

src/election_notification_works/voter_info.clj
```diff
@@ -79,10 +80,16 @@
...
+  response)
+
 (defn lookup [address & [google-election-id]]
   (let [address-string (util/map->string address [:street :street-2 :city
                                                   :state :zip])]
     (-> (google-civic-info-fetch address-string google-election-id)
+        log-on-error
```
mrichards42: Does the `response` include the original request? If not it would be helpful to also log the `address-string` and `google-election-id`

src/election_notification_works/voter_info.clj
eggsyntax: I was a bit hesitant to do warn since we don't know how many of these we'll be seeing. But either way's fine by me; I'll switch it.

src/election_notification_works/voter_info.clj
```diff
@@ -79,10 +80,19 @@
 (defn format-dropoff-locations [dropoff-locations]
   (map assoc-map-url dropoff-locations))
 
+(defn- log-on-error
+  [address google-election-id response]
+  (when-not (= 200 (:status response))
+    (log/warn "Error response from Google Civic API (for address"
+              address "and google-election-id" google-election-id "):"
+              response))
```
mrichards42: maybe `pr-str` this? That makes it easier to copy to a repl, plus it avoids the possibility of having a newline in the response (though unlikely here) which messes up the line of logging.

src/election_notification_works/voter_info.clj
```diff
@@ -79,10 +80,19 @@
...
+  response)
+
 (defn lookup [address & [google-election-id]]
   (let [address-string (util/map->string address [:street :street-2 :city
                                                   :state :zip])]
     (-> (google-civic-info-fetch address-string google-election-id)
+        ((partial log-on-error address google-election-id))
```
mrichards42: I don't _really_ care, but since I imagine we'll keep this logging around for a while, what about rearranging the args so you can make this `(log-on-error address google-election-id)` and have it take the response last?

Or even `(->> (log-on-error ...))` (though I'm not sure that's better than `partial`)? 🤷‍♂️  the double parens in the `partial` look pretty rough to me, but like I said, not a huge deal.

mrichards42: though agreed that a function like this would usually take the request last so you could do partial, so whatever looks best to you

src/election_notification_works/voter_info.clj
```diff
@@ -79,10 +80,20 @@
 (defn format-dropoff-locations [dropoff-locations]
   (map assoc-map-url dropoff-locations))
 
+(defn- log-on-error
+  [response address google-election-id]
+  (when-not (= 200 (:status response))
+    (log/warn
+     (pr-str "Error response from Google Civic API (for address"
+             address "and google-election-id" google-election-id "):"
+             response)))
```
mrichards42: oh, sorry I meant just the response, the parts that are definitely strings can be logged like normal

PR #175, opened 2019-02-27
add in-person deadlines to deadlines.registration

saramartinez:
This adds `in-person` and a deadline for `in-person` to voter registration methods and also adds a test for in-person deadline to the voting methods test.

This allows for the templates in `notification-works` (https://github.com/democracyworks/notification-works/pull/92) to remove hard-coded deadlines and instead use the in-person deadline if it’s listed on the election.

Pivotal: [Serve in-person deadlines in upcoming-election notys](https://www.pivotaltracker.com/n/projects/930636/stories/162861432)
Related PR: https://github.com/democracyworks/notification-works/pull/92


src/election_notification_works/deadlines.clj
```diff
@@ -21,6 +22,11 @@
         (merge (some-> reg-methods
                        (select-on-type :online)
                        (select-keys [:deadline-online])
+                       (apply-fn-to-mapvals coerce/to-local-date)))
+        (merge (some-> reg-methods
+                       (select-on-type :in-person)
+                       (select-keys [:end])
+                       (set/rename-keys {:end :deadline-in-person})
```
eggsyntax: Haven't run across `set/rename-keys`. Seems odd that it's in `clojure.set`, since it's fundamentally about maps (same with `set/rename` and `set/map-invert`). Not suggesting that you shouldn't use it, just interesting to discover that it exists :)

PR #177, opened 2019-03-01
Filter uninteresting errors; add test

eggsyntax:
Minor change from the previous PR


src/election_notification_works/voter_info.clj
```diff
@@ -80,13 +80,21 @@
...
   [response address google-election-id]
-  (when-not (= 200 (:status response))
-    (log/warn
-     (pr-str "Error response from Google Civic API (for address"
-             address "and google-election-id" google-election-id "):"
-             response)))
+  (when (nontrivial-error response)
+    (log/warn (pr-str "Error response from Google Civic API (for address"
+                      address "and google-election-id" google-election-id "):"
+                      response)))
```
mrichards42: While we're at it, let's only `pr-str` the data structures, not the strings. IDK if it makes more sense to log the address map, which you're doing here, or the address string that is sent to google. The map is more information, and you can always reconstruct the address string, but it is one more step.

PR #180, opened 2019-05-06
create context for all-mail counties and all-mail elections

li-at-dw:
Goes with the changes in 
https://github.com/democracyworks/election-works/pull/90
https://github.com/democracyworks/electorate-works/pull/71
https://github.com/democracyworks/notification-works/pull/91

For this ticket:  https://www.pivotaltracker.com/n/projects/930636/stories/163200093

Support notys for all-mail counties and all-mail elections by flagging voting-methods as `all-mail`, and then noting whether the election is all-mail due to `state`, `local`, or `election` rules.


src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -97,9 +98,13 @@
   preference for all-mail first, then excuse-required absentee, then
   no-excuse vote-by-mail, and if none of those are options, in-person."
   [electorate]
-  (let [voting-methods (get-in electorate [:district-division :voting-methods])]
+  (let [voting-methods (get-in electorate [:district-division :voting-methods])
+        primary-voting-method-source
+        (or (get-in electorate [:district-division :primary-voting-method-source])
+            :state)]
```
mrichards42: Is the `(or ... :state)` there for existing elections that don't have `:primary-voting-method-source`? If so I think it would make sense to do this part in `election-works`

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -97,9 +98,13 @@
...
     (condp some voting-methods
-      vm/all-mail? {:all-mail true}
+      vm/all-mail? {:all-mail true
+                    :all-mail-level {primary-voting-method-source true}}
```
mrichards42: Couple thoughts:

1. would it be helpful to have this as `primary-voting-method-source` (or whatever) for all voting methods? that could be a valid field for any voting method, right?
2. what do you think of `{:all-mail-level :state}` vs `{:all-mail-level {:state true}}`? the second reads a little strange to me for clojure, but it might make for a more readable template, since I think you'd have to do `{{#ifequals all-mail-level "state"}}` or something funny like that in the hbs template for the first option

eggsyntax: I personally prefer `{:all-mail-level :state}`; the latter is a bit jarring. `#ifequals` doesn't seem all that weird ;)

eggsyntax:
Obviously this isn't something for Li to do before she leaves, but...

Just from seeing this code and the previous PRs, it feels to me like the whole `primary-voting-method-source` approach is pretty awkward and abstract. I don't have any particular replacement in mind, but at some point I'd love to think about whether there's some less painful way to model that.

src/election_notification_works/handlers/upcoming_election.clj
li-at-dw: Yeah, I agree it's weird for Clojure.  I just went by the convention on the rest of the code, where we do stuff like {:all-mail true} {:excuse-required true}, and so on.  I wasn't quite sure if the handlebars syntax played well with keywords...

li-at-dw: Oh, but hey, yeah, I can just `#ifequals all-mail-level compare="state"`.  cool beans.

mrichards42:
@eggsyntax 

> it feels to me like the whole primary-voting-method-source approach is pretty awkward and abstract. I don't have any particular replacement in mind, and I've spent WAY less time thinking about this than y'all have, but at some point I'd love to think about whether there's some less painful way to model that.

My understanding is that we need to be able to distinguish a few cases:

1. all-mail states
2. all-mail local authorities (usually counties) within non-all-mail states
3. all-mail elections within non-all-mail states

Or put another way:

1. places where voting methods are the state defaults
2. places where voting methods are different from the state defaults
3. elections where voting methods are different from state defaults

eggsyntax:
> My understanding is that we need to be able to distinguish a few cases:

Is it the case that the correct set of voting methods is always the one with methods specified at the most local level (thinking of elections as the most "local")? That was my impression. If so, it _seems_ (without having tried to really analyze this) like there might be a simpler way to specify it, maybe by associating the voting methods with an attribute saying who they're specified by. Given that, I could imagine that everywhere in the code we would either have a) all applicable voting methods present, in which case we'd always be able to identify which is most-local, or b) have only the most-local set of voting methods, and be able to say from that which level they were at. I'm imagining this instead of having distinct attributes for the different sorts of voting methods.

This is all very vague speculation, though; I definitely haven't dug in at the level I'd need to to really think this through, whereas y'all have. It's really just intuition that there might be some simpler way of modeling this.

mrichards42:
> maybe by associating the voting methods with an attribute saying who they're specified by.

That's what `:primary-voting-method-source` is :)

Perhaps one more confusing thing is that data in `voting-methods` comes from two sources -- the election and the state or local authority. For the most part the election contributes different information (e.g. deadlines and dates) than the authority (which method is primary, which needs an excuse, which methods are supported). But an election can also override supported methods and the primary method.

So just `:voting-methods-source` won't cut it. The most accurate thing to do would model each attribute with `:value` and `:source`, but that's almost certainly overkill. So instead we've chosen to only care about the source of the `:primary` voting method.

mrichards42:
Not sure if this is also a source of confusion, but for clarity, it's the source of the primary voting method, not the primary source of voting methods

eggsyntax:
> That's what `:primary-voting-method-source` is :)

Ha! I do get that. I guess the above is a bit incoherent.

:man_shrugging: I'll let it go; I'm not clear enough on what I'm groping towards to come up with something useful. At some point if/when I'm dealing with this stuff in detail I'll give it some more thought.

PR #182, opened 2019-05-22
Prune dead turbovote signup

eggsyntax:
With help from Troy, identified some dead code that could be eliminated, left over from the transition away from a previous version of signups in which the context had to be rebuilt at every stage of user signup.

Have done some grepping, run unit tests, and gone through some signups, without any sign of problems. But if you can think of any problems that could be caused by these deletions let me know.


resources/config.edn
```diff
@@ -47,16 +47,6 @@
...
-    {:queue "election-notification-works.turbovote-signup.create"
-     :f election-notification-works.handlers.turbovote-signup/handle
-     :response true
-     :prefetch-limit #resource-config/edn #resource-config/env "TURBOVOTE_SIGNUP_PREFETCH"
-     :threads #resource-config/edn #resource-config/env "TURBOVOTE_SIGNUP_THREADS"}
-    {:queue "election-notification-works.turbovote-signup.delete"
-     :f election-notification-works.handlers.turbovote-signup/unschedule
-     :response true
-     :prefetch-limit #resource-config/edn #resource-config/env "TURBOVOTE_SIGNUP_UNSCHEDULE_PREFETCH"
-     :threads #resource-config/edn #resource-config/env "TURBOVOTE_SIGNUP_UNSCHEDULE_THREADS"}
```
mrichards42: 🎉 double-checked the 3 queues you deleted in datadog, and didn't even find metrics for them, so that says to me that they're completely unused

resources/config.edn
```diff
@@ -271,9 +256,9 @@
    :sms {:send-at {:hour 10 :minute 25}}
    :schedule-delay 10}
   :signup
-  {:schedule-delay 10}
+  {:schedule-delay 0}
   :subscription-confirmation
-  {:schedule-delay 2}}
+  {:schedule-delay 0}}
```
mrichards42: looks like some testing changes snuck in?

eggsyntax: Doh! Thanks for catching that.

mrichards42:
oh, and you should be able to get ride of the turbovote-signup endpoint in http-api as well

eggsyntax:
Not doing an immediate deploy to prod since no behavioral consequences.

PR #183, opened 2019-06-17
Welcome noty updates

eggsyntax:
Replaces the call to `registration-status.stream` with a call to `registration-status.page`, which returns the latest reg-status for the user. AFAICT, nothing is using the stream version now, and it could probably be eliminated (in `voter-registration-works`).


src/election_notification_works/handlers/signup.clj
```diff
@@ -75,10 +76,14 @@
...
+  "Sends a scheduled signup notification, populating user (including reg-status)
+  and subscriptions."
   [send-fn {:keys [user-id notification-type mediums]}]
-  (log/debug "sending any " notification-type " notys for " user-id)
-  (let [user (user/user user-id)
+  (log/debug "Sending any " notification-type " notys for " user-id)
+  (let [reg-status (select-keys (reg-status/latest-reg-status user-id)
+                                    [:method :status])
+        user (cond-> (user/user user-id)
+               reg-status (assoc :reg-status reg-status))
```
mrichards42: I'd prefer to avoid abbreviations in keys that we're going to send between functions and services. I think most places we spell this `:registration-status`.

Since this PR is called "Welcome noty updates", I'll assume you mean for reg status to end up in the welcome noty context. For that to happen, you need to do something with it [here](https://github.com/democracyworks/election-notification-works/blob/4a04d67302b3b031433b42aa51ebf08a28a3f740/src/election_notification_works/handlers/signup/email.clj#L48), and also somewhere in the sms context, thought the sms structure is a little more complex (though eventually I'd push for combining the sms and email contexts for simplicity's sake).

src/election_notification_works/handlers/signup.clj
```diff
@@ -87,7 +92,7 @@
                                (sub/read-subscription (dt/db))
                                :mediums
                                (into #{} filter-valid))]
-    (when (and (seq subscriptions) (not-ineligible? user-id))
+    (when (and (seq subscriptions) (reg-status/not-ineligible? user-id))
```
mrichards42: It would be nice if we didn't have to call voter-registration-works twice in the same function (since we already have a reg status), but I won't hold up the PR for it.

src/election_notification_works/handlers/signup.clj
eggsyntax: I don't think that latter is true -- it's coming through to `notification-status` with just these changes. I'm adding it to the user here, and the user is already sent along.

src/election_notification_works/handlers/signup.clj
eggsyntax: Yeah, I considered that. In the absence of time-constraints, I'd maybe change `reg-status/not-ineligible?` to take a reg-status instead of a user-id. But several other places call that function, and it seemed like a bad idea to slow things down by rewriting them all. I could just replace the above with the body of `not-ineligible?`, but then you lose the clarity of having the same fn called to assess eligibility. :man_shrugging:

mrichards42: Right, I was thinking the "short" version might be calling that function `not-ineligible-user?` and writing a second `not-ineligible-status?` function. But like I said, not a huge deal.

src/election_notification_works/handlers/signup.clj
eggsyntax: (& changed to `:registration-status`)

mrichards42: you mean `notification-send`?

Now that I look at this a little more, I think this is the wrong function to populated registration-status anyways, since this function is used for any signup-related notys (including the simple "welcome to turbovote" `:subscription-confirmation` noty you get as soon as you give us a phone number). If you need to add data to the context for the welcome noty specifically, it would go under the `signup.email` or `signup.sms` namespaces, which construct the context specifically for the welcome notys.

Not to say that this is the _right_ structure, but it is the structure that's currently in place.

gws:
All of the `with-redefs` in the tests which were stubbing out the `not-ineligible?` predicate need to change, since they aren't stubbing out the async channel anymore they cause the test suite to hang indefinitely.

eggsyntax:
> All of the `with-redefs` in the tests which were stubbing out the `not-ineligible?` predicate need to change, since they aren't stubbing out the async channel anymore they cause the test suite to hang indefinitely.

Yep, working on that now. Rushed this PR a bit too fast because I wanted to let someone else work on a piece, my bad :/

src/election_notification_works/handlers/signup.clj
eggsyntax: I considered that too, but I figured that it made sense to put it upstream from the ones specific to the sending method. Didn't catch that it was used for the `:subscription-confirmation`, though; I'll change it (probably just put it in `signup.email` for now since we may restructure/eliminate context for SMS.

gws:
Nothing to apologize for, just trying to save you some time!

test/election_notification_works/handlers/signup_test.clj
```diff
@@ -67,59 +68,70 @@
         subscriptions [:email :sms]
         expect-subscriptions (fn [expected]
                                (fn [{:keys [subscriptions]}]
-                                 (is (= expected subscriptions))))]
+                                 (is (= expected subscriptions))))
+        expect-reg-status-for (fn [uid] (expect-and-return
+                                        uid
+                                        {:method :online, :status :pending}))]
```
mrichards42: I don't think you need to stub out the reg status request here, just redeffing `user-not-ineligible?` to `(constantly true)` should suffice. It seems like `handle-send*` is only ever called with `expect-subscriptions` as its `send-fn`, so you shouldn't even get down to even `send-notifications`, much less `signup.email/send-signup*` where you actually fetch a real reg status.

src/election_notification_works/registration_status.clj
```diff
@@ -3,15 +3,36 @@
...
+            (let [status-msg (first statuses)]
+              (case (:status status-msg)
+                :ok (:registration-status status-msg)
+                :not-found nil
+                (throw (ex-info "Invalid status-message:" (pr-str status-msg)))))
+            (throw (ex-info "registration-status.page should only ever return a single reg-status"
+                            (pr-str reg-status-resp))))
+          (throw (ex-info "No reg-status found for user" user-id))))
+      (throw (ex-info "Error response from voter-registration-works"
+                      (pr-str reg-status-resp))))))
```
mrichards42: This chain of `if`s and `throw`s might be overly cautious. If this call succeeds, you're guaranteed to have 1 status per user-id, even if that status is `:not-found`

test/election_notification_works/handlers/signup/email_test.clj
```diff
@@ -104,9 +105,10 @@
 (deftest send-signup*-test
   (testing "full success"
     (with-redefs [util/signature (fn [] "All of us")
-                  util/user->local-auth-lookup-urls 
+                  util/user->local-auth-lookup-urls
                   (fn [u] {:local-reg-authority-lookup-url "http://reg.html"
-                           :local-election-authority-lookup-url "http://elec.html"})]
+                          :local-election-authority-lookup-url "http://elec.html"})
```
mrichards42: looks like existing indentation was changed here and a couple other places

eggsyntax: That's...very odd. It's correct in my editor, and I haven't changed it since these commits. Hard to see how it would then be incorrect on github. On the other hand, the only other one I see, in signup-test, *is* incorrect in my browser. :eyes:

test/election_notification_works/handlers/signup_test.clj
eggsyntax: Ah, yeah, thanks, nice catch. Left over from before I restructured adding reg-status into the recipient context. Fixed.

src/election_notification_works/registration_status.clj
eggsyntax: That's fair. I was feeling fairly paranoid when I wrote that :sweat_smile:

test/election_notification_works/handlers/signup/email_test.clj
eggsyntax: Wow, that sucks. I use a few lispy symbols in emacs, and in this case it was indenting as though `λ` was actually 1 char instead of 2. I don't think that's ever caused indentation problems for me before, but it sure is now :frowning_face: :frowning_face: :frowning_face: 
Fixed.

src/election_notification_works/registration_status.clj
mrichards42: This is still one level of paranoia too much for me, but I won't hold this PR up for it. You don't need this:
```clj
(if (pos? (count statuses))
  ...
  (throw (Exception. (str "No reg-status found for user" user-id))))
```

voter-registration-works is always going to send you back 1 status per user, so `(count statuses)` is always going to be 1. And if you _really_ don't trust it, you're still already throwing a (slightly) different exception when `(:status (first statuses))` isn't `:ok` or `:not-found`

eggsyntax: Good point! I'll get rid of that one too.

mrichards42:
Looks good, now that these tests have passed 👍

PR #185, opened 2019-06-26
Welcome noty updates 3

eggsyntax:
Adds the electorate (currently all of it) and the various reg deadlines & deadline types to the context sent to noty-works.

NB - stylistic improvement suggestions are welcome, but any non-blocking improvements will have to wait for a late PR because deadline.

NB - Solomon & Chris are working on the test that are failing.


src/election_notification_works/handlers/signup/email.clj
```diff
@@ -41,18 +44,46 @@
...
+  (when electorate
+    (let [date-formatter (get {:en util/en-email-date-formatter
+                               :es util/es-email-date-formatter}
+                              (:language-preference (:user electorate)))
+          format-date (fn [raw-date]
+                        (fmt/unparse date-formatter (coerce/to-date-time
+                                                     (coerce/to-date raw-date))))
+          deadline-map (deadlines/voter-reg-deadlines electorate)
+          online-deadline (format-date (:deadline-online deadline-map))
+          online-deadline-map (when online-deadline {:online-registration-deadline online-deadline})
```
mrichards42: What happens when you call `(format-date nil)`?

it might be more foolproof to do
```clj
(when online-deadline {:online-registration-deadline (format-date (:deadline-online))})
```

or even
```clj
(-> deadline-map
    (deadlines/apply-fn-to-mapvals format-date)
    (assoc :paper-deadline-type (some #{:deadline-postmarked :deadline-received} (keys deadline-map)))
    (set/rename-keys {:deadline-online :online-registration-deadline
                      :deadline-postmarked :paper-registration-deadline
                      :deadline-received :paper-registration-deadline}))
```

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -62,9 +93,12 @@
          :medium :email
          :language (:language-preference user)
          :context (merge {:recipient recipient
-                          :signers signers}
+                          :signers signers
+                          :electorate electorate}
```
mrichards42: I don't care that we're sending more data than we need here, but since we need to serve users who don't have an election with these notys, I think it would make more sense to pull out the reg and election authorities from the electorate, rather than send the whole thing.

```clj
{:election-authority (signup-util/authority->ctx
                      (or (:authority vbm-form-and-info)
                          (:election-authority electorate))
 :registration-authority (signup-util/authority->ctx
                          (or (:authority reg-form-and-info)
                              (:registration-authority electorate)
                              ;; the function leila's working on
                              (fetch-local-registration-authority user))
 :election (:election electorate)}

;; and remove the `:authority` logic from `->reg-form-ctx` and `->vbm-application-ctx`
```

Also, `authority->ctx` formats the authority name and gets the right address for you (mailing or physical address), so there's a little less logic needed in the template.

src/election_notification_works/handlers/signup/email.clj
teaforthecat: Yes, our changes were pretty much in line with these suggestions.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Definitely agreed that we need to fall back to @leilakass 's version once that's in place. Other than that, though, unless there are specific errors, I'd strongly prefer to not make unnecessary changes until we have a working version out there.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Cool, thanks!

src/election_notification_works/handlers/signup/email.clj
mrichards42: @eggsyntax I'm saying that stuff like [this](https://github.com/democracyworks/notification-works/blob/8ae6fbb8f0e03fa07351a1ea44bc0cc0324b0363/resources/notification-templates/turbovote-signup-confirmation-body.en.hbs#L28-L33) in the template isn't going to work. You always need to be able to have access to the registration-authority, even if there isn't an election.

eggsyntax: (You're totally right -- I'm just saying that once @leilakass ' code is ready, I agree that we need to add the reg-authority at the top level. I'm just proposing that we leave everything else as it is for now (& definitely swing back to it later once we have an initial working/testable version).

leilakass:
LGTM for this stage!

src/election_notification_works/handlers/signup/email.clj
mrichards42: I don't want to keep hammering on this, but I do want to make sure we're on the same page about where the data comes from.

if you don't have an election, but you asked to register by mail, we already have the reg authority -- it's just in `reg-form-and-info` -- and we don't need to wait for leila's code to have access to that for the most common case. Leila's code is important for the edge case when you _didn't_ ask to register by mail _and_ you don't have an election.

eggsyntax: No, this is *extremely* welcome!! I'm rushing and mistakes are all too plausible. But unless I'm badly mistaken, it's also there in the electorate (except in the edge case). Certainly it's coming through on the email signup noty that's being sent for me in krakenstein. I realize it's ridiculous to include it multiple times; I just want to get an initial version up for testing before going back and normalizing the context.

Am I missing something here?

mrichards42: RIght, both authorities will be on the electorate (except for the non-geocodable users, which are definitely edge cases). I'm just saying it's common that you wouldn't have an electorate.

eggsyntax: Ah, ok, I think I'm with you now. So rather than getting it from the electorate and falling straight back to Leila's, if we don't have an electorate we can try to get it from `reg-form-and-info` and only if we don't have _that_ do we need to resort to Leila's. Yah?

That's ignoring later refactoring where we ditch most of the electorate (and presumably get it from `reg-form-and-info` first when we have that.

For refactoring, my initial inclination without having dived much into it is to refactor away `->reg-form-ctx` and `->vbm-application-ctx`, or at least rename them, since the things they add aren't useful to just those cases anymore.

mrichards42: Right. I wrote the example preferring the form over the electorate, but the order of electorate vs form shouldn't matter.

eggsyntax: Lovely. Thanks for rephrasing that until I got it ;P

PR #186, opened 2019-06-26
[deadline] only show relevant deadlines

teaforthecat:
fixed both online and paper showing up when only online is present


src/election_notification_works/deadlines.clj
```diff
@@ -11,23 +12,49 @@
...
+    {:paper-registration-deadline-type type
+     :paper-registration-deadline (m type)}))
+
 (defn voter-reg-deadlines
+  "Returns a map of corrected deadline types with coerced dates
+  none or all of:
+  {:deadline-postmarked date
+   :deadline-received date
+   :deadline-online date
+   :deadline-in-person date}"
```
mrichards42: I like what you're doing with this function, but keep in mind that it's also used in the [upcoming election context](https://github.com/democracyworks/election-notification-works/blob/72a3a88dbf50d1167a69c6740648e140ec1db4b1/src/election_notification_works/handlers/upcoming_election.clj#L51), so we'll need to make sure it still works with that (in particular, with sms date formatters).

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -44,34 +44,21 @@
...
+  (let [formatter (date-formatter electorate)]
+    (fmt/unparse formatter (coerce/to-date-time
+                            (coerce/to-date raw-date)))))
 
 ;; TODO we're currently passing the entire electorate because of time pressure,
 ;; but we should use winnow out the tons of unneeded data when we can.
-(defn-traced send-signup*
+;; (defn-traced send-signup*
+(defn send-signup*
   "Given a user and associated info built in signup/send-notifications,
```
eggsyntax: This should still be `defn-traced`, unless I'm missing something.

eggsyntax:
Just in case you didn't notice, don't miss that the build is failing for some reason.

eggsyntax:
Just gave a very quick review, but offhand LGTM (pending build-fix, defn-traced). Thanks for adding all those tests!

src/election_notification_works/deadlines.clj
mrichards42: We also shouldn't change the name of these keys, since they're used in the [upcoming election template](https://github.com/democracyworks/notification-works/blob/welcome-noty-updates/resources/notification-templates/upcoming-election-body.en.hbs#L16-L18).

@eggsyntax It might be simpler to conform the new welcome noty template to use the existing deadline format rather than try to rewrite this function, e.g.
```hbs
Mail your completed form to your local election official using the pre-addressed, pre-paid envelope
{{#if deadlines.registration.deadline-postmarked}}
  by {{deadlines.registration.deadline.postmarked}}
{{else if deadlines.registration.deadline-received}}
  a few days before {{deadlines.registration.deadline-received}}
{{/if}}
. You must send your form to your local election official to complete the registration process.
We’ve included their address below:
```

This would also handle the case when we don't even have a reg deadline (b/c there isn't an election).

teaforthecat: That sounds good to me. I can try to make that change.

eggsyntax: Mike, are you saying this PR changes the name of those keys, or that `welcome-noty-updates-3` does?

If it's the former, @teaforthecat I request that this PR change the final behavior (ie the context produced by `send-signup*`) as little as possible -- your choices may be better ones, but I'm going to be merging changes from four different people, and the more folks make changes, the harder it's going to be to coordinate that and make the corresponding changes to the template.

That's only for this specific case, where there's time pressure and I'm starting to lose track of who (including me) is changing what...

eggsyntax: (and I'm perfectly happy to swing back around and make changes once we have an initial working version out there)

teaforthecat: That sounds good to me too. I will not try to make that change. I will keep the data structures the same (that has been my intention). I can get the tests passing once I've worked out the date type mismatches - moving the coercion into deadline.clj was a mistake.

eggsyntax:
Can't seem to approve because I've already left a top-level comment (anyone else run into this github weirdness?) but I'm happy with it given that tests are passing and if behavior hasn't changed. Thanks!

test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -433,7 +432,6 @@
 
       (let [message (:message (take-or-timeout channels/notification-send))]
         (is (= (-> message :context :election :date) "Saturday, February 1"))
-        (is (= (-> message :context :deadlines :registration :deadline-received) "Friday, January 17"))
```
mrichards42: let's leave these tests in

teaforthecat:
I pulled back on my changes and just made the tests pass.

test/election_notification_works/handlers/upcoming_election_test.clj
teaforthecat: done

eggsyntax:
> I pulled back on my changes and just made the tests pass.

I haven't looked at newest version, but just to be clear, I had no objection to refactoring, just changes to what ultimately went into the context. Feel free to add refactoring stuff back if you got rid of that. But I'll go ahead & merge this branch. Thanks for your flexibility under these (fortunately unusual) circumstances :)

PR #187, opened 2019-06-27
[deadline] add in-person and reg-deadline-and-type-test

teaforthecat:
This may be nice to have, I'm not sure, or it may fix a bug. In any rate, it confirms the behavior of reg-deadline-and-type.


eggsyntax:
LGTM on quick check. THanks!

PR #188, opened 2019-06-27
Welcome noty updates 4

eggsyntax:
Just adds state ID instructions.

Will look bigger than it is until welcome-noty-updates-3 is merged to master (diff to welcome-noty-updates-3 to review sooner).


src/election_notification_works/handlers/signup/email.clj
```diff
@@ -41,18 +45,80 @@
...
+
+(defn state-id-instructions
+  "Return the by-mail ID instructions for the upcoming election"
+  [recipient]
+  (let [state-ocd-id (->> (get-in recipient
+                                  [:addresses :registered :ocd-divisions])
+                          util/state-ocd-id)
+        reg-method (get-in recipient [:registration-status :method])
+        state-authority (authorities/get-state-authority state-ocd-id)]
+    (when (= reg-method :by-mail) ; online reg users have already seen these
```
mrichards42: While this is technically true that we only show these instructions to by-mail users, it doesn't hurt to include them.

I was talking about how we don't need the `:registration` instruction key from the `:online` voter-registraition-method.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -41,18 +45,80 @@
...
+      (-> state-authority
+          :ocd-division
+          last ; 3rd of the 3-element ocd-division vector
```
mrichards42: This is probably true, though it's not necessarily going to continue to be a 3-element vector. Probably safer to use `(nth 2)` since that's what we use everywhere else we need to do this.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -41,18 +45,80 @@
...
+(defn get-by
+  "Given a key and a value, gets the first element of sequence that matches"
+  [sq k v]
+  (first
+   (filter #(= (k %) v)
+           sq)))
+
+(defn state-id-instructions
+  "Return the by-mail ID instructions for the upcoming election"
+  [recipient]
```
mrichards42: Might be worth having this function take the state-reg-authority instead of the recipient, since you're going to need the state authority to get the other instructions (signature and citizenship instructions). Or perhaps having this be `state-reg-instructions` and return a map of all three instructions.

mrichards42:
Looks like this test is hanging -- is the state-reg-authority call you added in this PR already mocked in that test? Might be that.

eggsyntax:
> Would it work to base this PR off welcome-noty-updates-3 and then merge it into there instead of master? It would be a little easier to review, and it would be nice if we could have a single PR for the welcome noty context.

Can certainly frame it that way for now, at least, and I agree that ultimately merging a single PR is ideal. I've been carrying them one to the next for the sake of numerical order, but that does result in potential confusion. :man_shrugging:

src/election_notification_works/handlers/signup/email.clj
eggsyntax: I can rework that. Just finally looked at some production data, and what was throwing me off was that ID instructions in state authorities (for voting & reg) occur at:
```
voting-methods/in-person/id-instructions
voter-registration-methods/by-mail/instructions/id-number
voter-registration-methods/online/instructions/registration
```

...surely that can't be how we actually want to model it in the long term? Because that's wacky. But in the meantime, I can just follow a different path for `online` and `by-mail`. Not sure yet whether the oddball ones (`call-clerk`, `not-applicable`) ever have id instructions, guessing not, but in any case we don't show them in the welcome noty.

eggsyntax: To be a bit clearer, AFAICT we can't just follow `voter-registration-methods/by-mail/instructions/id-number` for `online`, so if we're not showing it, it may not be worth having a separate code branch to produce the online instructions. Maybe that's clear? Dunno, I'm a bit frazzled, gonna go stare at the ceiling.

mrichards42: The only instructions we need for the welcome noty are the by-mail reg instructions. Reg and voting methods also allow instructions, and they're used in different contexts (e.g turbovote-web, the cover letter if you get mailings, perhaps other notys?). The instruction keys are arbitrary, and 🤷‍♂ I guess we're using different keys in different places. which is a bit inconsistent.

Part of what might be confusing is that I keep insisting that you need 3 sets of instructions from the by-mail reg method (`:idnumber` `:signature` and `:citizenship-docs`). The done step uses all 3, so I was assuming that the welcome noty needs them as well. I just read through that second of the welcome noty and only noticed the id instructions, so sorry about making this more confusing.. I can double-check w/ Carmen to make sure leaving the other instructions off was intentional, since basically everything else about the welcome email parallels the done step text.

src/election_notification_works/handlers/signup/email.clj
mrichards42: Turns out I'm probably wrong about needing more than just id instructions, so ignore this.

src/election_notification_works/handlers/signup/email.clj
mrichards42: OK, so carmen says `:idnumber` and `:citizenship-docs` in separate paragraphs one right under the other. Turns out this is actually the same as what's on the done step, so I'm not _completely_ crazy.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -79,6 +80,30 @@
...
       (merge online-deadline-map paper-deadline-map in-person-deadline-map))))
 
+(defn get-by
+  "Given a key and a value, gets the first element of sequence that matches"
+  [sq k v]
+  (first
+   (filter #(= (k %) v)
+           sq)))
+
+(defn state-id-instructions
```
leilakass: Might be helpful to clarify in function name that this is for by-mail instructions. Everything else LGTM!

PR #189, opened 2019-06-27
Welcome voting context

mrichards42:
This isn't done, but @eggsyntax I'd at least like to have you aware of where I'm going w/ this.

Pieces to add to the context

* add the following keys to vbm-application:
  * ✅ too-late (for election date)
  * ~all-mail (election or state vms)~ not handling all-mail any more
    * ~all-mail-type (:state, :election, or :local)~
  * ✅ must-request (election or state vms)
* ✅ change election.authority to be top-level election-authority
* ✅ add recipient.voting-method-preference (added this to top-level context)
* ✅ add registration-status-link (from state authority)


mrichards42:
afaik this has been merged into the main WIP branch

PR #190, opened 2019-07-01
Welcome noty updates 6

eggsyntax:
There may be a handful of scenarios that aren't covered, notably geocoding fails, so this isn't ready for deployment, but I'd love to get this and the corresponding `notification-works` [PR](https://github.com/democracyworks/notification-works/pull/104) into master soon so we can start user testing on staging.
It could also use considerable cleanup and normalization of the context sent to noty-works, but that can wait since it's not a behavior change.


src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,149 @@
...
          :language (:language-preference user)
-         :context (merge {:recipient recipient
-                          :signers signers}
-                         auth-lookup-urls
-                         reg-form-ctx vbm-application-ctx)
+         :context (remove-nil-vals
+                   (merge {:recipient recipient ; break up. user name and state, :election-mail-subscription, :registration-status
+                           :signers signers
+                           :local-authority {:contact-info (select-keys local-reg-authority
+                                                                        [:physical-address :phone])}
```
mrichards42: I don't think we want `:local-authority` as a top-level key. We already have this data in `:registration-authority`. And even if we did, we'd want this to go through `authority->ctx`.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,149 @@
...
+        state-reg-method (matching-state-district-division-reg-method recipient state-authority)
+        reg-deadline-map (reg-deadline-and-type electorate)
         signers (util/signature)
         attachments (form-attachments reg-form-and-info vbm-form-and-info)
-        reg-form-ctx (->reg-form-ctx reg-form-and-info)
-        vbm-application-ctx (->vbm-application-ctx user vbm-form-and-info)
-        auth-lookup-urls (util/user->local-auth-lookup-urls user)]
+        vbm-application-ctx (or (->vbm-application-ctx user vbm-form-and-info)
+                                (default-vbm-ctx electorate user))
+        local-reg-authority (:authority (auth/local-reg-authority-search user))
```
mrichards42: This is always going to be an extra (uncached) rmq request, and at least most of the time, I don't think we need it. Could we do something like this so we only make the external service call when we need to?
```clj
(let [registration-authority (signup-util/authority->ctx
                              (or (:authority reg-form-and-info)
                                  (:registration-authority electorate)
                                  (:authority (auth/local-reg-authority-search user)))])
```

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,149 @@
...
                                 [:method :status])
         recipient (-> user
                       util/user->recipient
                       (assoc :election-mail-subscription mailings?
                              :registration-status reg-status))
+        state-ocd-id (->> (get-in recipient
+                                  [:addresses :registered :ocd-divisions])
+                          util/state-ocd-id)
+        state-authority (auth/get-state-authority state-ocd-id)
+        state-reg-method (matching-state-district-division-reg-method recipient state-authority)
```
mrichards42: I think basing state-reg-method off of the user's reg method makes things potentially more confusing and less flexible. The code that ultimately uses `state-reg-method` might work with the text we have in the welcome noty now, but I'm not sure it's in our best interest to have different contexts based on the user's reg method.

To illustrate, I believe these are the contexts that you might end up with:
```clj
;; by-mail reg users
{:ovr-url nil
 :state-instructions {:state-id-instructions "instructions ..."
                      :citizenship-docs "instructions ..."}}
;; online reg users
{:ovr-url "an actual url"
 :state-instructions nil}
```

What if we decide we want to include an ovr-url on, e.g. the "already registered" set of instructions? Now I doubt we need to show the state by-mail instructions for online reg users, but it still might be simpler to have functions like:
```clj
(defn ovr-url [state-authority]
  (-> state-authority
      (get-state-reg-method :online)
      :url))

(defn reg-instructions [state-authority]
  (-> state-authority
      (get-state-reg-method :by-mail)
      :instructions
      (select-keys [:idnumber :citizenship-docs]
      ,,,))
```

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,149 @@
...
+                                                                        [:physical-address :phone])}
+                           :election-authority (signup-util/authority->ctx
+                                                (or (:authority vbm-form-and-info)
+                                                    (:election-authority electorate)))
+                           :registration-authority (signup-util/authority->ctx
+                                                    (or (:authority reg-form-and-info)
+                                                        (:registration-authority electorate)
+                                                        local-reg-authority))
+                           :ovr-url (:url state-reg-method)
+                           :electorate electorate ; XXX
```
mrichards42: might be able to get rid of this now -- I believe all you need are election-authority, registration-authority, and election, which are now all on the context

src/election_notification_works/authorities.clj
```diff
@@ -56,14 +56,26 @@
...
+(defn local-authority-lookup-urls
   [state-authority]
   (select-keys state-authority
                [:local-reg-authority-lookup-url
                 :local-election-authority-lookup-url]))
+
+(defn local-reg-authority-search
+  [user]
+  (let [search-fn (wire-up/async->fn channels/local-reg-authority-search)
+        message {:ocd-divisions (into [] (get-in user
```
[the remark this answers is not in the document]
leilakass: it wasn't working before I added that!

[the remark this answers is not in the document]
mrichards42: Huh, I guess schema validates the type of collections? I guess that's one more plus for spec.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,149 @@
...
+      (merge online-deadline-map paper-deadline-map in-person-deadline-map))))
+
+(defn get-by
+  "Given a key and a value, gets the first element of sequence that matches"
+  [sq k v]
+  (first
+   (filter #(= (k %) v)
+           sq)))
+
+(defn matching-state-district-division-reg-method
```
leilakass: the name of this function is kind of confusing to me, is the "district-division" part hold meaning that I'm missing?

src/election_notification_works/handlers/signup/email.clj
leilakass: I agree with this thinking!^

src/election_notification_works/handlers/signup/email.clj
eggsyntax: I was trying to indicate that the reg-method doesn't belong directly to the state-authority, but to the state-authority's district-division. The multiple uses of reg-method and district division are confusing, and one of the things that slowed this work down for me. My mental model (at the domain level) of those different uses still isn't as good as I'd like it to be.

But now that you point out it's confusing, I think `matching-state-reg-method` would be better. Sound like a good choice?

leilakass: Yeah I agree! I think that the information about the reg-method belonging to the district-division belongs in the doc-string or some other documentation rather than the function name because the function itself is `state-authority->reg-method`

eggsyntax: Although I wouldn't tend to _name_ it `state-authority->reg-method` because it's doing something other than a transformation from a state-auth to a reg-method; it's taking both the state-auth and the recipient and finding the state-auth reg-method which _matches_ the recipient's reg-method.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Agreed, that's extra; I've got a note to eliminate it in the refactoring step.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Agreed!

src/election_notification_works/handlers/signup/email.clj
leilakass: Oo good catch! Thanks, I have a better understanding now

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -15,56 +21,147 @@
...
+        local-reg-authority (:authority (auth/local-reg-authority-search user))
+        state-authority-urls (util/user->state-authority-urls user)]
     (-> {:notification-type :turbovote-signup-confirmation
          :medium :email
          :language (:language-preference user)
-         :context (merge {:recipient recipient
-                          :signers signers}
-                         auth-lookup-urls
-                         reg-form-ctx vbm-application-ctx)
+         :context (remove-nil-vals
```
mrichards42: Do we need to remove nil vals? If it's important to remove them, is it also important to remove them recursively?

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Oh, yeah, silly of me; will fix.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: I suppose we don't really _need_ to. Typically I would take steps for each of several of the fns called to avoid merging the result into the context if the value is nil, but it seemed cleaner to eliminate them in one swell foop.

Offhand I don't remember seeing lower-level nils in this data, but I'll keep an eye out.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: I like it! I'll go that route.

eggsyntax: Seems reasonable to me! Changed.

src/election_notification_works/handlers/signup/email.clj
mrichards42: The one thing I'm (only slightly) concerned about with doing it at the end is that it might hide real errors. I agree that if there's a key with a `nil` value, that indicates a key that should have a value, and is an error. So it would serve as useful documentation if keys that should always be present could be found in the map that is the first argument to `merge` and keys that might not are in the rest of the maps.

test/election_notification_works/handlers/signup/email_test.clj
```diff
@@ -90,33 +58,142 @@
...
+                  util/user->state-authority-urls
                   (fn [u] {:local-reg-authority-lookup-url "http://reg.html"
-                           :local-election-authority-lookup-url "http://elec.html"})
-                  reg-status/latest-reg-status (fn [_] {:method :online, :status :pending})]
+                          :local-election-authority-lookup-url "http://elec.html"
+                          :voter-registration-status-url "http://status.html"
+                          :voter-registration-status-url-shortened "http://tvote.org/status"})
+                  reg-status/latest-reg-status (fn [_] {:method :online, :status :pending})
+                  auth/get-state-authority (fn [_] {:ocd-division
+                                                   ["ocd-division/country:us/state:mo"
```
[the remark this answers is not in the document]
eggsyntax: Gaaah. It's the damned fancy symbols acting up again. I really love them, and I hate to turn it off, but apparently I'm just gonna need to :cry: 

image

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Not a bad idea. I'll make a note to swing back around to this question on the refactor pass.

PR #191, opened 2019-07-02
Use TTL cache for election authorities

mrichards42:
... instead of LRU. See the discussion on #128

[Asana card](https://app.asana.com/0/1127682668253243/1127683140175785/f)


gws:
Thank you for doing this. I cannot thumb this up enough times. We maybe should have been more forceful about this back in #128, if only I'd thought of the argument outlined in the Asana card!

:+1: :100: :heart:

PR #192, opened 2019-07-09
Vote tomorrow changes

solomonsokoya:


test/election_notification_works/handlers_test.clj
```diff
@@ -593,3 +594,9 @@
       (is (= :ok status))
       (is (= {user-id-1 #{:email} user-id-2 #{:sms}}
              subscriptions)))))
+
+(deftest register-tomorrow-test
+  (testing "minimal happy path"
+    (let [test-electorate {:district-division {:voter-registration-methods [{:type :in-person
+                                                                             :end (tomorrow)}]}}]
+      (= true (util/register-tomorrow? test-electorate)))))
```
teaforthecat: can you move this so it is in the same namespace as the function please-and-thank-you

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -218,15 +218,18 @@
...
         tz (handler-util/get-tz-for-state state)
         send-after (calculate-send-after medium (:date election) tz)
         recipient-id (:id user)
         election-id (:id election)
-        language (:language-preference user)]
+        language (:language-preference user)
+        same-day-voter-reg (handler-util/same-day-voter-reg? electorate)]
     {:send-to "election-notification-works.vote-tomorrow.update"
      :send-after send-after
+     :same-day-voter-reg same-day-voter-reg
```
teaforthecat: can you add `:register-tomorrow (register-tomorrow? electorate)` please-and-thank-you

teaforthecat: so, in consider the comment below, the call to register-tomorrow? would look like `:register-tomorrow (register-tomorrow? electorate today)`

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -218,15 +218,19 @@
...
+     :register-tomorrow (handler-util/register-tomorrow? electorate (time/today))
```
teaforthecat: this is still going to make a system call for each message. Let's move the call to time/today into the function that calls this function `->vote-tomorrow-notification-messages` and pass it in. Then at least it is only getting called once per batch of messages. Also, maybe add a comment to that effect.

teaforthecat: oh, and now I see I misunderstood `->vote-tomorrow-notification-messages`, it is not a batch, it is only for each medium - so now I'm not sure what to do.

solomonsokoya: We are making a call to time/now in too-late-for-notification? and throughout this file (maybe we could use that for the today variable). Then maybe convert that and today variable using coerce/to-date and compare as such in register-tomorrow? ?

solomonsokoya: Nvm. I am wrong.

src/election_notification_works/handlers/util.clj
```diff
@@ -308,6 +308,24 @@
...
+  [electorate]
+  (->> (get-in electorate [:district-division :voter-registration-methods])
+       (filter #(= :election-day (:type %)))
+       seq
+       boolean))
+
+(defn register-tomorrow?
+  "If this is a vote tomorrow message, we can assume the user can register
+   tomorrow if the end of the in-person registration deadline is greater
+   than when this message is being sent."
```
eggsyntax: Could you add a bit to these comments and the ones on `same-day-voter-reg?` to clarify the distinction between them, ie (if I'm understanding them correctly) that this one means a voter can register tomorrow at the election office, and the other one means they can register at the polling place?

src/election_notification_works/handlers/vote_tomorrow.clj
teaforthecat: You are right that we seem to be making a system call via (t/now) for every upcoming-election message.

teaforthecat: I think we should put off the system call question for now, and leave it as it is.

eggsyntax: I would expect `t/now` to be lightweight enough that I wouldn't worry about it at this point; I'd say go for clarity over performance here, especially since notys are asynchronous.

PR #195, opened 2019-07-25
Rework and enrich email context to support welcome-noty changes

eggsyntax:
Could still use some refactoring and simplification of both the code and the message sent to notification-works, but that needs to wait till after the preliminary deploy.

Note simplified/squashed commit history :wink:


src/election_notification_works/handlers/signup/email.clj
```diff
@@ -134,27 +171,37 @@
...
-         :identifiers {:recipient-id (:id user)}}
-        (merge (when (seq attachments) {:attachments attachments}))
-        send-fn)))
+        state-authority-urls (util/user->state-authority-urls user)
+        ;; We place the next-contact-section in any of several places:
+        contact-section-location (cond
+                                   (or all-mail-state?
+                                       (= reg-method :in-person)) "after-reg-section"
+                                   mailings? "after-voting-section"
+                                   :else "after-printer-section")
```
mrichards42: Not sure I understand the rationale for doing this, but if nothing else . . . if you have mailings, you won't have an after-printer section, so aren't those last two conditions effectively the same?

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -21,14 +22,45 @@
...
+      (when (.isAfter date now)
+        date))))
+
+(defn lang->date-formatter
+  "Takes a language keyword (:en or :es), returns a date formatter"
+  [lang]
+  (get {:en util/en-email-date-formatter
+        :es util/es-email-date-formatter}
+       lang
+       util/en-email-date-formatter))
```
mrichards42: `case` might be easier to read

test/election_notification_works/handlers/signup/email_test.clj
```diff
@@ -4,10 +4,18 @@
...
             [election-notification-works.voting-methods :as vm]
             [clj-time.core :as time]
+            [clj-time.coerce :as coerce]
             [clojure.test :refer :all]
             [election-notification-works.test-helpers :refer [random-uuid]]
             [election-notification-works.handlers.util :as util]))
 
+(defn future-date
+  "reg-deadline-and-type filters out dates in the past, so we need to test it
+  with future dates. Pass a number of days ahead (defaults to 90)."
```
mrichards42: fyi I think the standard way to deal w/ this in tests is to use [`clj-time.core/do-at`](https://github.com/clj-time/clj-time/blob/c6711ebd6b471038446867c3c053152882464eca/src/clj_time/core.clj#L773)

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -21,14 +22,45 @@
...
   "There may be useful information even if no form was generated, ie
   states where users must request the form."
   [user vbm-form-and-info]
   (when (seq vbm-form-and-info)
-    (let [must-request (boolean (and (:wants-vbm? vbm-form-and-info)
+    (let [formatter (lang->date-formatter (:language-preference user))
+          must-request (boolean (and (:wants-vbm? vbm-form-and-info)
                                      (not (:has-form? vbm-form-and-info))))
-          deadline (signup-util/->date user (:deadline vbm-form-and-info))]
+          deadline (format-date (inst-if-in-future (:deadline vbm-form-and-info))
```
mrichards42: If this is a `java.util.Date`, and `format-date` says it takes a `clj-time` date, how does this work?

Might spare a little headache by converting this to a clj-time date and then using `date-if-in-future` (which would also allow you to get rid of `inst-if-in-future` I believe).

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -21,14 +22,45 @@
       voter-reg-form-uri (conj (assoc pdf-attachment :url voter-reg-form-uri))
       vbm-appl-form-uri (conj (assoc pdf-attachment :url vbm-appl-form-uri)))))
 
+(defn format-date
```
mrichards42: I'd consider renaming this `format-future-date` and doing the "is this in the future" test right here.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: :man_shrugging: direct request from Carmen, not sure what motivated it.

Good point -- I'll circle back around with Carmen and figure out what she was thinking of re the relationship between those two conditions.

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Yeah, unifying dates/datetimes is high on my list of stuff to deal with post-deploy.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -21,14 +22,45 @@
...
+  [inst]
+  (when (= 1 (compare inst (coerce/to-date (time/now))))
+    inst))
+
+(defn- date-if-in-future
+  "Takes a clj-time date; returns that date if it is a future date, or nil"
+  [date]
+  (when date
+    (let [now (.toLocalDate (time/now))]
+      (when (.isAfter date now)
```
leilakass: (for my own learning) what does the syntax for `.toLocalDate` and `.isAfter` mean?

leilakass:
LGTM!

src/election_notification_works/handlers/signup/email.clj
eggsyntax: Sure! `clj-time`, the library we off use for date/time stuff, is a wrapper around the JodaTime Java library, so we're dealing with Java objects here. `(.some-method some-object)` means: call the `some-method()` method on the object `some-object`. So it's the same as what would be written (in Java & many other OO langs) `some-object.some-method()`. Following items are argumnets, so in the case of `isAfter`, it's the equivalent of `date.isAfter(now)`.

leilakass: Okay cool that was really helpful, thanks for explaining!

PR #196, opened 2019-07-25
[upcoming] add reg-deadline-passed

teaforthecat:


src/election_notification_works/handlers/util.clj
```diff
@@ -308,26 +308,49 @@
...
+  [electorate]
   (let [methods (get-in electorate [:district-division :voter-registration-methods])
+        election-date (coerce/to-date-time (get-in electorate [:election :date]))
         in-person (first (filter #(= :in-person (:type %)) methods))]
-    (if-let [end (and in-person (:end in-person))]
-      ;; must have the same type to compare end is a java.util.Date
-      (time/before? (coerce/to-date-time today) (coerce/to-date-time end)))))
+    (if-let [end (and in-person (coerce/to-date-time (:end in-person)))]
+      (or (time/before? election-date end)
+          (time/equal? election-date end)))))
```
mrichards42: If this used `coerce/to-local-date` instead of `to-date-time`, could this just be `equal?` instead of before or equal?

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -164,6 +168,10 @@
                         :election election
                         :election-authority election-authority
                         :same-day-voter-reg same-day-voter-reg
+                        :reg-deadline-passed (reg-deadline-passed? electorate)
```
mrichards42: This needs to go in `->updated-notification-message` since it's based on the send date, and `->upcoming-notification-message` is called when the message is scheduled, which is likely well before it's sent.

src/election_notification_works/handlers/util.clj
```diff
@@ -308,26 +308,49 @@
...
+(defn reg-online?
+  "can register online"
+  [electorate]
+  (let [methods (get-in electorate [:district-division :voter-registration-methods])
+        online (first (filter #(= :online (:type %)) methods))]
+    (some? online)))
+
+(defn reg-same-day?
+  "Returns true if the last day to register in-person is the same as the election
+  day. This is for the vote tomorrow notification."
```
mrichards42: Back to the conversation we had in the notification-works PR -- this function is really `reg-same-day-at-office?` since it's based only on the in-person reg deadline. I definitely think this is the function you need, but I think it needs the additional clarification that we're only looking at the `:in-person` reg method, not the `:election-day` reg method.

Since this function is already `(and reg-same-day reg-at-office)`, I'm back to thinking that you don't need to explicitly check both of those in the template.

teaforthecat: I think I'll add a check for :election-day to reg-same-day? to keep the concepts separate. You are probably write, but it feels better to keep them separate.

mrichards42: If you base `reg-same-day` on both :election-day reg method and :in-person :end date, that creates the following problem:
```clj
(let [reg-at-polling-place true
      reg-in-person true
      election-day = #inst "2019-11-05"
      reg-in-person-end #inst "2019-11-01" ;; before election day
      ;; this is what I assume you mean by "add a check for :election-day"
      reg-same-day (or reg-at-polling-place
                       (equal? election-day reg-in-person-end))]


  ;; Assuming you keep the order the same in the template, you're OK...
  (hbs/render
   "{{#if reg-same-day}}
     {{#if reg-at-polling-place}}
       You can register at the polls on election day.
     {{else if reg-in-person}}
       You can register at your election office through election day.
     {{/if}}
   {{/if}}")
  ;; "You can register at the polls on election day."

  ;; ... but if you switch the order
  (hbs/render
   "{{#if reg-same-day}}
     {{#if reg-in-person}}
       You can register at your election office through election day.
     {{else if reg-at-polling-place}}
       You can register at the polls on election day.
     {{/if}}
   {{/if}}")
  ;; "You can register at your election office through election day"
  ;; that's not true
  )
```

src/election_notification_works/handlers/upcoming_election.clj
teaforthecat: moved

src/election_notification_works/handlers/util.clj
```diff
@@ -308,26 +308,49 @@
   (time/latest (coerce/to-date-time send-at)
                (coerce/to-date-time (minutes-from-now delay))))
 
-(defn same-day-voter-reg?
-  "Determine if an election allow same-day voter registration. If true voter the voter
+(defn ^:deprecated same-day-voter-reg?
```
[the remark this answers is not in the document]
teaforthecat: done

src/election_notification_works/handlers/util.clj
teaforthecat: I'm not familiar enough with the domain, but I'd rather take the cautious option here and check for both.

src/election_notification_works/handlers/util.clj
teaforthecat: Ok I'll remove the check for :election-day.

src/election_notification_works/handlers/util.clj
mrichards42: Deadlines are election-specific, so you'll never see deadlines after and election. That makes this code a little confusing to me (since we're checking for a case that indicates an error), but it's technically correct, so your call.

src/election_notification_works/handlers/util.clj
```diff
@@ -308,26 +308,42 @@
...
+  (let [methods (get-in electorate [:district-division :voter-registration-methods])
+        in-person (first (filter #(= :in-person (:type %)) methods))]
+    (some? in-person)))
+
+(defn reg-online?
+  "can register online"
+  [electorate]
+  (let [methods (get-in electorate [:district-division :voter-registration-methods])
+        online (first (filter #(= :online (:type %)) methods))]
+    (some? online)))
```
mrichards42: `reg-at-polling-place?`, `reg-at-office?`, and `reg-online?` are all checks for the existence of reg method. Might be worth writing a `get-reg-method` function that these are al based on.

src/election_notification_works/handlers/util.clj
```diff
@@ -308,26 +308,42 @@
...
        seq
        boolean))
 
-(defn register-tomorrow?
-  "If this is a vote tomorrow message, we can assume the user can register
-   tomorrow if the end of the in-person registration deadline is greater
-   than when this message is being sent. To sum it all if true the user can register
-   to vote tomorrow at the election office."
-  [electorate today]
+(defn reg-at-office?
```
mrichards42: any reason to call this `reg-at-office?` instead of `reg-in-person?`, which is what we use everywhere else?

teaforthecat: You could also say that you are registering in-person at the polls, at least a new person could, so this is a nice distinguishing term.

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -123,6 +123,10 @@
                      (not (get-in context [:recipient :should-request]))))]
     (assoc-in context [:recipient :paper-vbm-form] status)))
 
+(defn reg-deadline-passed? [electorate]
+  (every? #(time/before? (coerce/to-date-time %) (time/now))
+          (vals (deadlines/voter-reg-deadlines electorate))))
```
mrichards42: Looking at the `deadlines/voter-reg-deadlines` code, I think this is checking online, by-mail, and in-person deadlines, but I believe you only want to be checking online and by-mail, since in-person is handled differently.

teaforthecat: I'm not seeing the concern here sorry, do you mind clarifying? How is in-person handled differently?

mrichards42: With this code you're never going to hit `reg-deadline-passed and reg-same-day and reg-at-office` (i.e. [this part of the template](https://github.com/democracyworks/notification-works/blob/5129be228b846dab3dd9cc1db29f7941aee7a27d/resources/notification-templates/upcoming-election-body.en.hbs#L27)).

If the in-person deadline is election day, it will never be in the past, so `reg-deadline-passed` will be false.

mrichards42: And given the language that's in the template ("The deadline to register online or by mail") I assume we only want to check those two deadlines.

teaforthecat: I mean I think it is ok to check the :end date more than once because `reg-deadline-passed?` should be false if :online has passed but :in-person has not.

teaforthecat: oh, duh, plus it is right in the template text "by mail or online has passed"

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -123,6 +123,19 @@
...
 
+(defn reg-deadline-passed?
+  "warn the user that they can not use mail or online, but they may be able to
+  register in-person or on election-day at the polls"
+  [electorate]
+  (let [now (time/now)
+        methods (get-in electorate [:district-division :voter-registration-methods])
+        by-mail (first (filter #(= :by-mail (:type %)) methods))
+        online (first (filter #(= :online (:type %)) methods))
+        postmarked (:postmarked by-mail)
```
mrichards42: `by-mail` can be `:postmarked` or `:received`

teaforthecat: Are there cases of `:received` without `:postmarked` ? Would we need different messaging depending on the attribute? Maybe in this case we would use `:received` as a fallback and simply say that it is too late to reg by mail even though it was too late two days ago or whatever. That way, there is a gap, but we can't say what we don't know, and we don't know how long the mail will take.

mrichards42: Yes, reg deadlines are always either `:received` or `:postmarked` -- we never have both, so it's not that one is a fallback for the other.

What you had before might have been simpler, since the logic for extracting deadlines is already there in `deadlines/voter-reg-deadlines`. It would just be a matter of whitelisting the deadlines you want to check:
```clj
(defn reg-deadline-passed? [electorate]
  (let [reg-keys #{:deadline-postmarked :deadline-received :deadline-online}]
    (->> (deadlines/voter-reg-deadlines electorate)
         (filter (comp reg-keys key))
         (every? #(time/before? (val %) (time/now))))))
```

teaforthecat: true, but let's wait on that improvement.

teaforthecat:
this passed the test on staging, ok to merge?

PR #199, opened 2019-08-06
Voting method preference fix

eggsyntax:
We were (incorrectly) using the voting-method-preference from the vbm-form, and hence showing vbm instructions (& attaching vbm form) even for users who said they were voting at the polls; this corrects that. It also does some consolidation around voting-method-preference, and ensures that the call to voting-methods to get it can be easily replaced in tests.


src/election_notification_works/handlers/signup/email.clj
```diff
@@ -165,15 +162,17 @@
         all-mail-state? (all-mail? state-authority)
         reg-deadline-map (reg-deadline-and-type electorate)
         signers (util/signature)
+        voting-method-preference (voting-method-preference-fn (:id user))
```
mrichards42: ... or maybe it's simplest to leave what you have changed (especially since it's clearer code), but make this

```clj
(if (:wants-vbm? vbm-form-and-info)
  :by-mail
  (voting-method-preference-fn (:id user)))
```

eggsyntax: What about just including `:voting-method-preference` in the map returned by `vote-by-mail-works.handlers/make-application-form` along with `:wants-vbm?`? Presumably that shouldn't break any other users of `make-application-form`, and that way we could eliminate the call to `voting-method-works.voting-method-preference.read` entirely. It's unfortunately non-obvious that `vbm-form-and-info` should be the source of the user's voting method preference, but yay for fewer external calls...

eggsyntax: Either way I'm going to go ahead & merge this PR just to get the corrected behavior into prod, but I'll go back and make one or another of these changes during my refactoring pass.

mrichards42: sure, that works

eggsyntax: Looks like it isn't used anywhere else anyway :)

PR #200, opened 2019-08-08
Welcome sms 2

leilakass:
A lot of these changes are simply matching the new changes that Egg made in a branch for email welcome notys. I'm reusing the same code for the email welcome notys for the sms welcome notys so that we can eventually refactor the code to use the same function for building the context. I deleted SMS tests that weren't relevant anymore, will still need to add others in.


mrichards42:
Looks like we have a pretty big merge conflict on this. I'm assuming we can leave all the email code alone and just merge code that's changed in:
* `handlers.signup.sms` -- the main changes in this PR
* `handlers.signup.sms-test` -- removing all the tests
* `handlers.signup` -- making the call to `sms/send-signup` work like it does with emails
* `handlers.signup-test` -- tests for the above changes

If that's the case, I think the easiest way to get there would be creating a new branch off the current master (which would already have all the email changes merged in), and then copy all your sms changes over. Then we can make a single commit that deploys this to staging and changes the `:schedule-delay` -- I'd prefer if those were together instead of mixed in w/ the other commits, since we can revert that one before the whole thing is merged.

I'd be happy to work with you on that if you'd like.

PR #201, opened 2019-08-13
Welcome noty tech debt

eggsyntax:
Does a bit of cleanup and refactoring. I'm poised-ish to do a final step of simplifying/normalizing the context (and not sending so much unneeded stuff) but I'm leaving that until after `welcome-sms` is done, since we're planning (at least tentatively) to unify the context generation between those two (the sms one is currently a modified copy of this one).


eggsyntax:
Tracking down build failure.

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -22,26 +22,14 @@
...
-
-(defn- date-if-in-future
-  "Takes a clj-time date; returns that date if it is a future date, or nil"
-  [date]
-  (when date
-    (let [now (.toLocalDate (time/now))]
-      (when (.isAfter date now)
-        date))))
+    (let [datetime (coerce/to-date-time (coerce/to-date raw-date))]
+      (when (.isAfter datetime (time/now))
```
mrichards42: why not `time/after?` to keep it clojurey?

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -22,26 +22,14 @@
       voter-reg-form-uri (conj (assoc pdf-attachment :url voter-reg-form-uri))
       vbm-appl-form-uri  (conj (assoc pdf-attachment :url vbm-appl-form-uri)))))
 
-(defn format-date
-  "Takes a clj-time date & clj-time date-formatter; returns a formatted datestring"
+(defn format-future-date
```
mrichards42: If we're doing a bunch of refactoring, it might actually be cleanest if we _didn't_ format dates before sending them to the template, and let the template figure out formatting. That would be one less difference between sms and email templates . . . something like:
```clj
  (defhelper fmt-date [ctx options]
    (let [format-str (hbs.helper/hash options "format")
          locale (hbs.helper/hash options "locale")
          formatter (cond-> (fmt/formatter format-str)
                      locale (fmt/with-locale (java.util.Locale. locale)))]
      (fmt/unparse formatter (coerce/to-date-time ctx))))
```

And then we could have global partials
```hbs
;; en-email-date.hbs
{{fmt-date this format="EEEE, MMMM dd"}}
;; es-email-date.hbs
{{fmt-date this format="EEEE, MMMM dd" locale="es"}}
;; sms-date.hbs
{{fmt-date this format="M/d"}}
```

which would be used in a template like:
```hbs
{{> en-email-date deadline-registered}}
{{> es-email-date deadline-registered}}
{{> sms-date deadline-registered}}
```

src/election_notification_works/handlers/signup/email.clj
eggsyntax: If only that existed :wink:

src/election_notification_works/handlers/signup/email.clj
```diff
@@ -22,26 +22,14 @@
...
-    inst))
-
-(defn- date-if-in-future
-  "Takes a clj-time date; returns that date if it is a future date, or nil"
-  [date]
-  (when date
-    (let [now (.toLocalDate (time/now))]
-      (when (.isAfter date now)
-        date))))
+    (let [datetime (coerce/to-date-time (coerce/to-date raw-date))]
```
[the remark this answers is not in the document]
eggsyntax: I would have sworn there was something we were using that `coerce/to-date` alone would trip over. But if so, I can't repro it now. So :+1:

src/election_notification_works/handlers/signup/email.clj
eggsyntax: That's a really decent idea IMO. Not going to try to do it at this point because Leila & I have been keeping things tightly in sync, and this would be a relatively complicated one to coordinate. But might be a really good one for later!

PR #202, opened 2019-08-13
Welcome sms 3

leilakass:
-using updated email welcome noty code for sms noty
-updated the date format used for SMS (will refactor with email date formatting later)
-set shorter noty send time and allow for "mailme" hack for testing
-fix failing tests and remove old SMS noty tests


src/election_notification_works/handlers/signup/sms.clj
```diff
@@ -1,125 +1,193 @@
...
+  [date]
+  (when date
+    (let [now (.toLocalDate (time/now))]
+      (when (.isAfter date now)
+        date))))
+
+(defn ->vbm-application-ctx
+  "There may be useful information even if no form was generated, ie
+  states where users must request the form."
+  [voting-method-preference user vbm-form-and-info]
```
eggsyntax: The `user` argument is unused now, but I like your decision to keep it since we're hoping to unify these. I expect that in the end we'll want to add a `medium` argument so that logic can vary on email vs sms -- unless we go with Mike's idea of moving the date formatting to the templates, at least.

src/election_notification_works/handlers/signup/sms.clj
```diff
@@ -1,125 +1,193 @@
...
+                   msg (-> {:notification-type :turbovote-signup-confirmation
+                            :medium :sms
+                            :language (:language-preference user)
+                            :context context
+                            :identifiers {:recipient-id (:id user)}})]
+               (send-fn msg)))
 
 (def send-signup
-  (partial send-signup* notification/send))
+  (partial send-signup* notification/send vm/voting-method-preference))
```
eggsyntax: Just went through the line-by-line comparison with `signup/email` -- looks like there are few enough differences that it might be marking the bits that there are (for ease of merging them later). I wish I'd kept track while I went through it, but the ones I remember are (I think) the `:reg-form-uri` and `:vbm-form-uri` lines, the differences in `->vbm-application-ctx`. and the formatter in `reg-deadline-and-type`. I might be forgetting some though, or mixing them up with the most recent changes in `email` that haven't been incorporated here.

Just a thought, though; certainly not a necessity!

eggsyntax: Ha! Now I see that Mike made the same comparison.

leilakass:
Yes, the only changes are 
1. using the sms date formatting
2. adding the form uris to the context instead of treating them as attachments

So that's what we have to keep in mind when we refactor with email code!

PR #208, opened 2019-09-14
[upcoming] add early-voting url to context

teaforthecat:


teaforthecat:
@mrichards42  can you take a look at this test coverage? Should I add any scenarios?

mrichards42:
@teaforthecat will give this a look after NVRD and by the end of the week, assuming nothing crazy happens tomorrow

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -56,16 +56,22 @@
...
-                         util/es-email-date-formatter
-                         util/en-email-date-formatter)]
-    (when-let [early-voting (deadlines/early-voting-deadlines electorate)]
-      (deadlines/apply-fn-to-mapvals early-voting
-                                     (partial fmt/unparse-local-date formatter)))))
+  (let [methods (get-in electorate [:district-division :voting-methods])
+        early-voting (first (filter #(= :early-voting (:type %)) methods))]
+    (some-> early-voting
+            (update :start (formatter language))
+            (update :end (formatter language)))))
```
mrichards42: It looks like the [early-voting schema](https://github.com/democracyworks/election-works/blob/914454829a2d34a90e3dce626ddb32b6e5740f1f/src/election_works/entities/election_schema.clj#L37-L38) doesn't require `:start` or `:end`. It's likely that we always provide them, but I'm not completely sure.

So maybe we need this?
```clj
(cond-> early-voting
  (:start early-voting) (update :start (formatter language))
  (:end early-voting) (update :end (formatter language)))
```

src/election_notification_works/handlers/upcoming_election.clj
```diff
@@ -56,16 +56,22 @@
      {:date (some->> vbm-deadline
                      (fmt/unparse-local-date formatter))}}))
 
+(defn formatter [lang]
+  (partial fmt/unparse-local-date
+           (if (= lang :es)
+             util/es-email-date-formatter
+             util/en-email-date-formatter)))
```
mrichards42: Would you consider sticking a function like this in `util` next to the date formatters? Seems like we should be using this logic anywhere we format a date. Or maybe even this logic w/ a medium argument, since we also have an sms formatter.

We could also stick a `clj-time.coerce/to-local-date` in there as part of the fn, which is a no-op for joda-time dates, and would deal w/ java.util.Dates

test/election_notification_works/handlers/upcoming_election_test.clj
```diff
@@ -961,3 +961,18 @@
                                                                                   :deadline-online %}]})]
       (is (= false (reg-deadline-passed? (electorate tomorrow))))
       (is (= true (reg-deadline-passed? (electorate yesterday)))))))
+
+(deftest early-voting-test
+  (testing "basic happy path"
+    (let [electorate {:district-division {:voting-methods [{:type :early-voting
+                                                            :url "http://url"
+                                                            :start (yesterday)
+                                                            :end (tomorrow)}]}}
```
mrichards42: Couple things here: (1) I believe the input data uses `java.util.Date`s, not joda-time dates, and (2) if we're just testing that these exist as strings in the output, could we use static dates?

teaforthecat:
You are raising some valid concerns @mrichards42, we should probably hold off on deploying this until after nvrd.

PR #211, opened 2019-09-24
Vote tomorrow send time

mrichards42:
[Asana card](https://app.asana.com/0/1127682668253243/1140757185671283/f)

I almost took this as an opportunity to tear the vote-tomorrow noty down and start from scratch, but in the end I think this is good enough for now. The test code is a real mess (mostly b/c it's relying on async stubs) and I didn't bother to change many of the fundamentals, except to make sure that it still works. I also added a couple tests to insure that (a) we have an election-id in the initial scheduled message, and (b) we don't send messages for users w/o an electorate (similar to upcoming-election code).

I'd still like to re-work the whole noty, but more likely after we build out a standardized way to create contexts for election notys as part of [this card](https://app.asana.com/0/1127682668253243/1140758582607570/f).


src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -259,36 +258,40 @@
...
-  "Takes an incoming `notification`, looks up the polling locations
-  and adds one of them to the context. The `channel` is used to send out
-  the new notification."
-  [{{:keys [recipient]} :context, :as notification}]
-  (let [updated-notification (update notification :context update-voter-info)]
-    (notification/send updated-notification)
-    {:status :ok
-     :notification updated-notification}))
+  "Takes an incoming `notification` with cached VIP lookup data, creates the
+  notification context, and sends the notification."
```
eggsyntax: What's the VIP lookup data? That's not ringing any bells for me (I haven't had much to do with VIP). & is it worth validating that it's present in the context?

src/election_notification_works/handlers/vote_tomorrow.clj
```diff
@@ -259,36 +258,40 @@
...
-                 (pr-str notifications))
-      (doseq [notification notifications]
-        ;; this has to mimic what is in
-        ;; notification-works.handlers.notification/schedule-notification->scheduled-message-works-format
-        (let [identifiers (merge (select-keys notification
-                                              [:medium :notification-type])
-                                 (:identifiers notification))]
-          (handler-util/unschedule-notification identifiers))))))
+  (doseq [medium mediums]
+    (let [;; we only care about identifiers, so create a message with an empty context
```
eggsyntax: Really helpful clarification, thanks! :pray:

src/election_notification_works/handlers/vote_tomorrow.clj
mrichards42: That's optional, actually -- it's what we get from the civic info api (I used "vip" here, since it's what the fns are called already).

In fact, it's the reason this one ended up being more complicated. This code builds the noty at schedule time, and then overwrites (most of) it at send time. The `update-voter-info` fn, which hits the civic-info api, relies on the structure of the noty context. We should change that at some point, but I didn't want to get into a huge refactor on this PR.

